PR #40, opened 2016-02-01
Fixed problem with filters

juniwalk:
Trying to assign value from persistent holder to uninitialized filter.

### DIFF
--- a/src/DataGrid.php
+++ b/src/DataGrid.php
@@ -835,6 +835,10 @@ class DataGrid extends Nette\Application\UI\Control
 	public function assableFilters()
 	{
 		foreach ($this->filter as $key => $value) {
+			if (!isset($this->filters[$key])) {
+				continue;
+			}
+
 			if (is_array($value) || $value instanceof \Traversable) {
 				if (!ArraysHelper::testEmpty($value)) {
 					$this->filters[$key]->setValue($value);


paveljanda:
Thanks.

juniwalk:
Looks like this bug was there because I switched o5/grido to your datagrid and grido probably uses same scheme for "filters". Explains why I saw the problem immediately after installing the datagrid.

paveljanda:
Yes. This error occurred also when you did some filtering and then removed a used filter from grid.

juniwalk:
Oh I see, did I fix it properly then?

paveljanda:
Almost yes. :)

There still remains the value in session, I didn't notice that before - thanks again. I don't know if it is a good idea to unset it.

... Yes, it should not be there. You may create a new filter with different values to select. Then the session value won't be in current select range.

I will implement that immediately and release version 1.0.4.

juniwalk:
Alright :) I am having another issue with tree rendering, I will debug it and report to you back later.

JakubKontra:
Is that bug about nodes moving?

juniwalk:
@JakubKontra It tries to read `has_children` value from my entity as if it was array, which of course fails horribly.

paveljanda:
That should not happen. The only place, where it is reading that property is the tree view template on line 16 and there is a check whether given item is an object or an array:

```php
{var $has_children = is_object($item) ? $item->{$tree_view_has_children_column} : $item[$tree_view_has_children_column]}
```

juniwalk:
I am aware of that, I saw it in tracy report, trange thing is, I just tried it again and it works. :D

PR #46, opened 2016-02-03
Sortable - renamed parameter 'id' to 'item_id'

DavidZadrazil:
When you have for example renderDetail($id) and you are recieving GET parameter $id from that render method, signal handleSort($id, ...) does not work right. So I renamed **id** parameter in JS files to **item_id**. 

Precompile standalone.min.js is needed.


paveljanda:
Thanks for PR.

PR #47, opened 2016-02-04
Row: Fixed getting row Id using internal value getter

juniwalk:
This is a huge problem as `$id` property of my entity is private so getter method needs to be used.


paveljanda:
Jeez, i forgot this! Thank you!

But please, do not use the harcoded 'id' column/property name, but the `$primary_key` instead.

juniwalk:
Already ahead of you, fixed :)

Issue #58, opened 2016-02-07
Add option for custom name of handle for sorting


martinknor:
Because if you have more than one grid on page - you cannot handle it :)

paveljanda:
A changed the method and property name a bit. And also added the same to tree_view.latte.

PR #59, opened 2016-02-07
Add option for custom name of handle for sorting

martinknor:
#58


paveljanda:
Once there is a possibility to set custom handler, i would prefer setting handler including the 
exclamation mark. Can you change that please?

PR #78, opened 2016-02-10
fix notice when IDataSource obj. is directly set into DataGrid

### LINKED ISSUE #77: Notice  Undefined variable: data_source
When I send into `DataGrid` datasource directly (Api annotation allows that) I got that notice -> exception.

**How to reproduce**:
```php
/**
	 * @return DataGrid
	 */
	public function createComponentGrid()
	{
		$grid = new DataGrid();
		$grid->addColumnNumber('id', 'Id')
			->setSortable()
			->setFormat(2);

		$queryBuilder = $this->queryBuilderProvider->getQueryBuilder();
		$grid->setDataSource(new DoctrineDataSource($queryBuilder, 'id'));

		return $grid;
	}
```

**Cause** (In: `DataGrid::setDataSource($source):` 
```php
# here in this case nothing happens actually after that if-else three

if ($source instanceof DataSource\IDataSource) {
			// $source is ready for interact
} else ...
...
...

# After if-else three $data_source remains undefined
$this->dataModel = new DataModel($data_source);
....

Achse:
fixes #77

### DIFF
--- a/src/DataGrid.php
+++ b/src/DataGrid.php
@@ -136,7 +136,7 @@ class DataGrid extends Nette\Application\UI\Control
 	 * @var bool
 	 */
 	protected $sortable = FALSE;
-	
+
 	/**
 	 * @var string
 	 */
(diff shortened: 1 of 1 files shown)


Achse:
Duplicate

PR #81, opened 2016-02-10
attempt to add some proper annotation mostly for return static in fluent

Achse:
If return (mostly `@return static`) is not annotated correctly, IDEs and code sniffers have a lot of yapping about it.

### DIFF
--- a/src/DataGrid.php
+++ b/src/DataGrid.php
@@ -136,7 +136,7 @@ class DataGrid extends Nette\Application\UI\Control
 	 * @var bool
 	 */
 	protected $sortable = FALSE;
-	
+
 	/**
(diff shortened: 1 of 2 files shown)


src/DataGrid.php
```diff
@@ -1751,6 +1768,7 @@ public function getItemsDetail()
 	/**
 	 * Items can have thair detail - toggled
 	 * @param mixed $detail callable|string|bool
+	 * @return Column\ItemDetail[]
```
Achse: Why setter even returns something?

paveljanda: Because of that fluent interface. `$grid->doThat()->andDoThis();`

PR #83, opened 2016-02-10
Added possibility to have grid in component and call its handlers in addAction

Achse:
problem:

```php
class XyzGridControl extends BaseControl
{

     ....
     $grid->addAction('confirm', 'Schválit', 'confirm!');
}
```
causes: `Unknown signal 'confirm', missing handler XyzModule\XyzPresenter::handleconfirm()`

----

Question is: Can this change broke something?

### DIFF
--- a/src/Column/Action.php
+++ b/src/Column/Action.php
@@ -94,11 +94,11 @@ class Action extends Column
 		}
 
 		$a = Html::el('a')
-			->href($this->grid->getPresenter()->link($this->href, $this->getItemParams($row)));
+			->href($this->grid->getParent()->link($this->href, $this->getItemParams($row)));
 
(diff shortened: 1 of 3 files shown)


paveljanda:
Great!

Yes, it can, in that 100 downloads. If someone is using it "incorrectly", it would be a BC break. How should i decide? How to implement this and don't break back compatibility?

Achse:
* One option is to release it with BC break as new version.
* Another option is catch exception and call `getPresenter()` instead as fallback.

paveljanda:
Could you please add the second solution? I thought about it too.

It will be removed in next version.

Achse:
Sure, working on it.

PR #87, opened 2016-02-11
Temporary hotfix of annotation for setDataSource

Achse:
It ugly I know but at this time its best I can offer.


paveljanda:
Thanks. I will make it more readable at the weekend.

PR #93, opened 2016-02-15
Added some black magic for getting correct translator

Achse:
I do not like this, but I have no idea how make this clean-way.

What do you think?

The problem is that you have to write:
```php
$grid = new DataGrid();
$grid->setTranslator($this->getTranslator());
```

Any time in order to have correct translator instance in DataGrid component.

### DIFF
--- a/src/DataGrid.php
+++ b/src/DataGrid.php
@@ -1595,7 +1595,15 @@ class DataGrid extends Nette\Application\UI\Control
 	public function getTranslator()
 	{
 		if (!$this->translator) {
-			$this->translator = new Localization\SimpleTranslator;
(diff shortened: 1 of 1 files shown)


paveljanda:
The clean way is to create some DataGridFactory, which you would inject into your Presenters (other components). This factory would set each datagrid your translator:

```php
class DataGridFactory
{
    $grid = new DataGrid;
    $grid->setTranslator($this->translator);

    return $grid;
}
```

I do not like this either, sorry. :)

juniwalk:
@paveljanda :+1: not a good kind of a magic to get into code.

PR #95, opened 2016-02-15
Added setDisable($callback) on Action

Daaarkling:
Hi, I added this new functionality into Action. It allows to disable (not render) action, if given callback function will return TRUE (user may not have permission to see this action).

Hope it helps.

### DIFF
--- a/src/Column/Action.php
+++ b/src/Column/Action.php
@@ -55,6 +55,11 @@ class Action extends Column
 	 * @var array
 	 */
 	protected $confirm;
+	
+	/**
+	 * @var callable|NULL
+	 */
+	protected $disable_callback;
 
 
 	/**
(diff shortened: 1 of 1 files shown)


paveljanda:
How does that differ from http://ublaboo.paveljanda.com/datagrid/row > Row condition?

You can allow/forbid some of your actions or group action rendering.

Daaarkling:
I missed that one. But still can you give me example how to do it, forbid rendering of spesific action (only delete action for example) for specific row. Thanks.

paveljanda:
Exactly that is in the documentation i sent you. :)

```php
$grid->allowRowsAction('delete', function($item) {
	return $item->id != 3;
});
```

Daaarkling:
Great thx. I just didn't see that. Ignore my PR then :-)

PR #98, opened 2016-02-16
DataGrid: Make the table stripped (colored even rows)

juniwalk:


paveljanda:
I had it there, but have to remove.. Do you know what will happen if you are using item detail? :)

juniwalk:
Hmm, what will happen? Maybe add `$grid->addClass('table-striped');` then?

paveljanda:
You can try that yourself. :) The striped-schema breaks. Item detail has to be loaded into another row (`<table>/<tbody>` can not have some other children than <tr>) and that will switch all remaining event <-> odd table rows color schema.

paveljanda:
Maybe there is a solution although not so nice. Item detail can be loaded into 2 table rows.  The second will stay empty/hidden/whatever. Just to let the striped color schema work.

paveljanda:
I will look at it.

juniwalk:
Alright, I will look into it once I've got some spare time.

paveljanda:
DataGrid is now by-default "striped", table class can be chaged in {block #table-class} when extending parent template

juniwalk:
I've noticed, great work! :)

PR #104, opened 2016-02-21
Method orderBy used in DoctrineDataSource::sort changed to addOrderBy

Daaarkling:
Hi,
with doctrine datasource, if I use $grid->setDefaultSort(['foo1' => 'ASC', 'foo2' => DESC]) only last item in array is used for sorting. I found out that in DoctrineDataSource::sort method is orderBy but it should be addOrderBy. Only hope that this not break anything else :-)


JakubKontra:
+1 @paveljanda

PR #120, opened 2016-03-01
OnColumnAddCallback

harmim:
- add $onColumnAdd callback called when you add column of any type to datagrid

I need call event when I adding columns in my grid. Is there any other way to do it? If not can you plese merge it?

### DIFF
--- a/src/DataGrid.php
+++ b/src/DataGrid.php
@@ -227,6 +227,9 @@ class DataGrid extends Nette\Application\UI\Control
 	 */
 	protected $can_hide_columns = FALSE;
 
+	/** @var callable[] */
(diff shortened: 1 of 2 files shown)


paveljanda:
I can add this. Can you please just change the namespaces a bit? In use statement above the `DataGrid` class is: `use Ublaboo\DataGrid\Column`, co you can remove the `use Ublaboo` and write this:

```php
protected function addColumn($key, Column\Column $column)
```

instead of

```php
protected function addColumn($key, Ublaboo\DataGrid\Column\Column $column)
```

Sorry for being so pedantic.

harmim:
It´s changed.

Issue #129, opened 2016-03-08
Tree-view: param $tree_view_has_children_column as callback


vaclavpavek:
Hello, it would be good if the second parameter could be callback.

paveljanda:
Right. Can you send a PR? :)

PR #131, opened 2016-03-09
tree view children column as callback

vaclavpavek:
PR for issue #129 Tree-view: param $tree_view_has_children_column as callback.

This PR not tested, but extract from exists project


paveljanda:
I will just slightly modify it

vaclavpavek:
Nice :-), have a nice day

PR #156, opened 2016-03-23
Add possibility to create group action input box..

Attanon:
I want to add group actions with input box. For group changes for example prices. Please review this and comment.

### DIFF
--- a/src/DataGrid.php
+++ b/src/DataGrid.php
@@ -1364,16 +1364,36 @@ class DataGrid extends Nette\Application\UI\Control
 
 
 	/**
-	 * Add group actino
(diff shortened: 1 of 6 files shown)


Attanon:
I change it little bit. No it has no BC breaks.

Attanon:
?

paveljanda:
Thanks, I will at it tomorrow and let you know. So far i like that.

paveljanda:
@Attanon Can you please remove the `@deprecated` annotation? I think there can be an alias that won't throw an exception.

Attanon:
Changed

PR #170, opened 2016-04-07
Unite reset filter string when there are no rows found

juniwalk:
Was there a reason for that `ublaboo_datagrid.here` anchor?

### DIFF
--- a/src/templates/datagrid.latte
+++ b/src/templates/datagrid.latte
@@ -258,7 +258,7 @@
 							<tr n:if="!$rows">
 								<td colspan="{$control->getColumnsCount()}">
 									{if $filter_active}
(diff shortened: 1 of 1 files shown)


grongor:
I looked into the same just now - I think the author intended to use it like this https://github.com/ublaboo/datagrid/pull/171

paveljanda:
Actually, for the fist time it was something like that:

No items found, you can reset the filter <a>here</a>.

So neither of these 3 options is perfectly correct, but all can fit the usage..

juniwalk:
@paveljanda I see, how shall we proceed?

paveljanda:
I would leave it like it is now. It would not mean any possible BC break.

PR #171, opened 2016-04-07
Use correct translation string

grongor:
The string `ublaboo_datagrid.no_item_found_reset` wasn't used anywhere so I guess this is where it belongs :-)

### DIFF
--- a/src/templates/datagrid.latte
+++ b/src/templates/datagrid.latte
@@ -258,7 +258,7 @@
 							<tr n:if="!$rows">
 								<td colspan="{$control->getColumnsCount()}">
 									{if $filter_active}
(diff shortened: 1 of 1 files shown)


grongor:
So you will leave it without any change? That doesn't seem right.

paveljanda:
It is explained here: https://github.com/ublaboo/datagrid/pull/170

Do you have any other solution that seams better?

grongor:
Well I think that this is a bug and therefore BC shouldn't be a priority there and it should be fixed. It shouldn't be left like that.

paveljanda:
Right. I forgot there is also almost same sentence stored under the same key.

grongor:
Thanks ;-)

PR #181, opened 2016-04-13
Support for components from callback

romanmatyus:
With this feature it's possible attach component, for example for *item detail*.

```php
$grid->components['orderDetail'] = function () use ($grid) {
	return $this->orderDetailFactory->create();
};
```

Now it's possible use in `detail.latte`:

```latte
{control orderDetail $item->id}
```

### DIFF
--- a/src/DataGrid.php
+++ b/src/DataGrid.php
@@ -77,6 +77,11 @@ class DataGrid extends Nette\Application\UI\Control
 
 	/**
 	 * @var array
+	 */
+	public $components = [];
+
+	/**
+	 * @var array
 	 * @persistent
 	 */
 	public $filter = [];
@@ -2164,6 +2169,18 @@ class DataGrid extends Nette\Application\UI\Control
 		}
 
 		return $this->items_detail;
+	}
+
+
+	/**
+	 * Factory for components from callback
+	 * @return Nette\ComponentModel\IComponent
+	 */
+	protected function createComponent($name)
+	{
+		return (isset($this->components[$name]))
+			? $this->components[$name]()
+			: parent::createComponent($name);
 	}
 
 


paveljanda:
It is possible. But you don't have much control over receiving the data and manipulation with rendering after submitting the form.

I plan putting something like that directly into ItemDetail.

romanmatyus:
Ok..
Off topic.. It's possible to have more then one different details in one row - with different icons and content?
For example detail of order, detail of customer, etc.

paveljanda:
No it is not ... by now.

paveljanda:
See https://github.com/ublaboo/datagrid/commit/a12f4a9402f7377b6f231e860b1b41a41fd3c7b1

romanmatyus:
Hello, was added support for forms in detail, but still it's not possible use in template:

```latte
{control invoice $item->id}
```

Have my solution some problem or it's wrong approach?

For example, I have component for displaying invoice in detail of grid. Component is same as in another places in app. How it's now possible show detail?

Thanks.

romanmatyus:
In ideal case will be callback for component in format:

```php
$this->components_callback['invoice'] = function (DataGrid $grid, $row = NULL /* returned by IDataSource::filterOne()*/) {
};
```

Btw... My fault, parameter `$components` it's already in DataGrid from class [Nette\ComponentModel\Container](https://api.nette.org/2.4.0/source-ComponentModel.Container.php.html#$components).

paveljanda:
@romanmatyus The thing is, it is absolutely important for `DataGrid` to have control over it's components. In your case, `DataGrid` can not take responsibility for possible breaks in for example forms validation (related: https://github.com/ublaboo/datagrid/issues/197).

Issue #184, opened 2016-04-14
DataGridPaginator - translator change


pavelgajdos:
Move $this->template->setTranslator($this->translator); from render() method to createTemplate method. (quicker solution, not the best one, keep reading..)

This would allow setting translator of the paginator component later than during the instance creation.

The issue is that when I assign a new translator using the decorator extension to my grid instance, this new translator is not set to the paginator component as I would expected. I tried to do a workaround, calling $this['paginator']->getTemplate()->setTranslator($translator) when decorator is fired, but it has no effect as the translator of the paginator component is set in the constructor and then assigned to the template in render method of the paginator component. The best case scenario would be to adjust setTranslator of the grid object to call setTranslator on paginator component (the latter method would have to be created).

paveljanda:
I don't see, why is that necessary.

How it works now:

1, You create a datagrid object
2, You set all columns, filters, translator
3, When template is rendered, paginator component factory is called
4, If you set previously some translator to datagrid, it is passed also to paginator in __constructor

Your example should work. Can you please show me the config (the decorator setting)? I don't see any reason why that should not work

pavelgajdos:
Problem solved. I was accessing $this['paginator']->getTemplate() sooner than I should (in method 'attached' of the grid instance) because of other settings. This caused the paginator template to be created before my translator was set to the grid instance. I moved this settings to onRender event of the datagrid and now it works just fine. Thank you for clearing things up. It just seemed to me that the decorator is invoked too late.

PR #185, opened 2016-04-14
Setting a new translator to datagrid instance causes the translator t…

pavelgajdos:
…o be set to all sub-components that implement method 'setTranslator'. Solves #184


paveljanda:
Why is that? `DataGridPaginator` component gets translator instance in constructor (`DataGrid::createComponentPaginator()`).

pavelgajdos:
Irrelevant. See #184.

PR #194, opened 2016-04-21
Translate confirm message before replacing value

VaclavSir:


VaclavSir:
Sorry, I didn't check current master, this is already fixed by 11d02986.

paveljanda:
Yes it is fixed in master. But you know, i can use the tests. :)

PR #200, opened 2016-04-26
Add button Show default columns

romanmatyus:
I have very big grid - with 50 hideable columns. When I click to *Show all columns*, it's very lengthy manual click on every column for setting *default view*.
This PR add *Show default columns* button.
image


paveljanda:
I like that!

Thanks.

PR #207, opened 2016-05-04
composer: works with Nette 2.4

dg:


legendik:
@paveljanda Can you update packagist? On packagist is still require `~2.3.0`

PR #218, opened 2016-05-10
Hide group action stuff if there's no group action available on rows

dakorpar:
I think we should not have Group actions select box, and select all checkbox if group actions are enabled, but in all displayed rows there's no group action available. So I created this pull request, I'm very new to this datagrid, so it's highly possible that I overseen something.


dakorpar:
I was few commits behind so there were few conflicts, I'll do it again from scratch.

PR #220, opened 2016-05-10
Fix error: Column in where clause is ambiguous

romanmatyus:
Fix empty table name for columns, for example when I get one row for detail and datasource have JOIN.

snimka obrazovky z 2016-05-10 23 45 04


paveljanda:
You should use `$grid->addColumnText('id', 'Id', 'item.id');`. It is not an datasource error, it is user error.

romanmatyus:
Thanks, but correct answer it's: Use `$grid->setItemsDetail(TRUE, 'car.id');`

*It's realy user (my) error.*

PR #222, opened 2016-05-11
Hide group action checkboxes and selectbox if group action is not all…

dakorpar:
animation
 
With this fix is just -> 
screenshot_2

### DIFF
--- a/src/DataGrid.php
+++ b/src/DataGrid.php
@@ -384,13 +384,13 @@ class DataGrid extends Nette\Application\UI\Control
 		}
 
 		$callback = $this->rowCallback ?: NULL;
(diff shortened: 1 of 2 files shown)


paveljanda:
Can you please make both `TRUE` and `FALSE` constants uppercase?

dakorpar:
Yes, sorry :)

dakorpar:
ignore last two commits, it should be another pull request, but  I accidentelly created new branch from origin/master instead of ublaboo/master. I did it wrong when doing this  thing in master in first place...

paveljanda:
Ehh. Can you please squash these commits into one?

PR #224, opened 2016-05-12
Fixes sorting when database column is specified and is different name…

dakorpar:
If you set sorting by database column instead of column key (you have to when you have join query) and have remember state on grid when trying to get sessiong back it tries to read grid column which doesn't exist and you get exception

Code that reproduces this bug:
`
        $this->setDefaultSort(['column.id' => 'DESC']);
`
refresh the page 
and you'll get exception: "There is no column at key [$key] defined."

### DIFF
--- a/src/DataGrid.php
+++ b/src/DataGrid.php
@@ -1508,6 +1508,9 @@ class DataGrid extends Nette\Application\UI\Control
 		 */
 		if (empty($this->sort_callback) && !empty($this->sort)) {
 			foreach ($this->sort as $key => $order) {
+				if(!isset($this->columns[$key])){
+					continue;
+				}
 				$column = $this->getColumn($key);
 
 				if ($column && $column->isSortable() && is_callable($column->getSortableCallback())) {


paveljanda:
When you have a join a outputting the data like this:

```
$grid->addColumnText('name', 'Name', 'table2.name');
```

You simply set the default sort via column key:
```
$grid->setDefaultSort(['name' => 'asc']);
```

paveljanda:
Key of the array in argument of `DataGrid::setDefaultSort();` is a column key, not the column of database table.

dakorpar:
@paveljanda  This 
`$grid->addColumnText('name', 'Name', 'table2.name'); `
would not work, you would have to set 
`$grid->addColumnText('name', 'Name', 'name');` 
'table2.name' does not exist you would get notice: User Notice

Attempt to read missing column, and column would be empty on grid
and if you have name in join and main query ofc your query would tell you 
`Column 'name' in order clause is ambiguous`

paveljanda:
@dakorpar Of course column `table2.name` does not exist in your schema.

Show me your datagrid factory and datasource preparation.

dakorpar:
no, you're correct, I have no idea what I was thinking, to much problems latelly :)

PR #225, opened 2016-05-12
Hide group action checkboxes and selectbox if group action is not allowed on any row

dakorpar:
animation
 
With this fix is just -> 
screenshot_2


src/DataGrid.php
```diff
@@ -384,15 +384,24 @@ public function render()
...
+				$hasGroupActionOnRows = TRUE;
+			}
+			
 			if ($callback) {
 				$callback($item, $row->getControl());
 			}
 		}
 
+		if($hasGroupActionOnRows){
+			$hasGroupActionOnRows = $this->hasGroupActions();
```
paveljanda: Why that line?

dakorpar: The problem is that `$row->hasGroupAction()` returns true if no group actions are defined, so this checks at the end if group actions are actually there since $hasGroupActionOnRows will be true and  grid maybe does not have group actions at all. If we skip this then it's possible to have screwed up grid (checkboxes on rows, and no group actions at all)

Solution was either that or to go with new loop if group actions are available, thought that it's going to be faster with one loop only, although many grids does not contain group actions at all.
if you want I can refactor so that if group actions exists we do a new loop and set it there, probably more readable option....

paveljanda:
What if someone has perfect styles for particular datagrid table and suddenly a table header and one column disappears?

Arrgh, i like it, but it could mean broken datagrid table appearance. I will leave it here for a bit longer, until i decide which approach is better.

paveljanda:
Can you please resolve the conflicts? I will merge it then.

paveljanda:
@dakorpar Thanks

PR #227, opened 2016-05-16
Added option to summarize numeric columns

juniwalk:


juniwalk:
One big problem I see with this is that you have to set items per page to ALL, or it won't show summary for whole table but just for current page. (Which can be feature)

juniwalk:
#228 - Possible source of conflicts if merged.

PR #230, opened 2016-05-19
typos, removed non-breaking spaces

dg:


dg:
It is interesting that they appeared in the code :-)

paveljanda:
It is indeed.

I will build both `.js` and minified versions later this day.

PR #231, opened 2016-05-20
Change outer filter class (outer div and all outer filters)

Attanon:
I just need to change the class of outer div in outer filter. And all outer filters. So I created PR with added block to change that.


paveljanda:
The datagrid template should not be altered just because of added class. Please use your own template.

PR #232, opened 2016-05-20
Display column summary only when there are any rows

juniwalk:


paveljanda:
Can you please add `empty($rows)` instead of array -> bool type casting?

juniwalk:
@paveljanda It is done :)

PR #237, opened 2016-05-23
Changed DataMode.phpl to first call getData() before getCount()

jakubvojacek:
The grid does always know what page is currently set and what is the item-per-page count, right? Therefore we can tell the datasource this (`limit(offset, itemsperpage)`) before asking for getCount().

I did test it on my grids and it worked well, you think this could work or it could break something?

### DIFF
--- a/src/DataModel.php
+++ b/src/DataModel.php
@@ -111,14 +111,15 @@ class DataModel
 		 */
 		if ($paginator_component) {
 			$paginator = $paginator_component->getPaginator();
-			$paginator->setItemCount($this->data_source->getCount());
 
 			$this->data_source->sort($sorting)->limit(
 				$paginator->getOffset(),
 				$paginator->getItemsPerPage()
 			);
 
-			return $this->data_source->getData();
+			$data = $this->data_source->getData();
+			$paginator->setItemCount($this->data_source->getCount());
+			return $data;
 		}
 
 		return $this->data_source->sort($sorting)->getData();


paveljanda:
How can you call
```php
$this->data_source->sort($sorting)->limit(
  	$paginator->getOffset(),
  	$paginator->getItemsPerPage()
  );
```

when $paginator is not set?

paveljanda:
Sorry, man.

jakubvojacek:
Why does paginator need to know item data count? The grid knows on what's the current page as well as how many items should be displayed on page (https://github.com/ublaboo/datagrid/blob/master/src/DataGrid.php#L2111-L2112). Thats all that is needed to calculate offset. 

Or why the `$paginator` should know the itemcount from datasource before calculating offset?

paveljanda:
For example, `$paginator->getOffset()` behaves differently in situation `$item_count` is given and `$item_count` is not given. Just look in the code.

Is there any value added on your PR?

jakubvojacek:
It solves https://forum.nette.org/cs/25444-ublaboo-datagrid-mocny-rychly-rozsiritelny-hezky-anglicky-dokumentovany-datagrid?p=7#p174685

The problem was that in order to `getCount()`, the data source would have to query the API (when using `ApiDataSource`) which is redundant when the API is capable of returning the result & count in just one call. Using the aforementioned solution I was able to change that behavior and load the grid with just one API call.

To fix the the problem you mentioned (different behavior of `Paginator::getPageIndex()`) I could do  something like this (first get data & count, then compare result of getPageIndex and if not same, refetch):
```
/**
 * Paginator is optional
 */
if ($paginator_component) {
    $paginator = $paginator_component->getPaginator();

    $original_page_index = $paginator->getPageIndex();
    $this->data_source->sort($sorting)->limit(
        $paginator->getOffset(),
        $paginator->getItemsPerPage()
    );

    $data = $this->data_source->getData();
    $paginator->setItemCount($this->data_source->getCount());
    
    if ($paginator->getPageIndex() !== $original_page_index) {
        // not item_count is set, lets compare the original page index with new one, if there's diffence, lets fetch the data again
        $this->data_source->sort($sorting)->limit(
            $paginator->getOffset(),
            $paginator->getItemsPerPage()
        );
        // refetch data with new offset
        $data = $this->data_source->getData();
    }
    return $data;
}
```

Or do you have any other idea how to resolve this? Because when there are 5 grids on one page, it's obviously better to do just 5 api http request instead of 10 (as those are quite time consuming compared to db queries).

PR #244, opened 2016-05-25
Now you can set up default per page value, not the first from the list

dakorpar:
I have some requests on my project to set up default per page value to something else then first value, I thought it would be nice to have in datagrid as well so I created PR


paveljanda:
Why won't you just set different order items per page in the select? (40 - default, 20, 1000, 'all')

And why the `PHP_INT_MAX `?

dakorpar:
@paveljanda because it would look wierd to have 
- 50
- 10
- 25
- 100

In my opinion this option should be available.
I was NOT happy at all with PHP_INT_MAX
but if I've set per_page to 'all' then it would always be displayed only 1 item, when changing in grid to all then it would be good. I saw that when creating paginator I have to set items per page as number, I cannot return null there, so I've put PHP_INT_MAX, although number of 1000 should probably be more then enough since more then that would ussually even get out of memmory exception, or would even crash the server.

paveljanda:
If you look deeper into the code of `DataGrid` class, you will see (`DataGrid::getPaginator()`), that in case whn 'all' is selected paginator behaves like it does not exist. Yes it would mean high page load, but that is up to the programmer. Please remove the PHP_INT_MAX..

One other thing - please make it separate method, just like `DataGrid::setDefaultSort()`.

dakorpar:
Found it, fixed, thanks for info, did not understand that code back then, I'm still new to this awesome Datagrid :))

paveljanda:
Thanks, i will modify it just a bit and release in next version.

PR #259, opened 2016-06-08
Filter DataSource:  \DateTime check when using  \DateTime::createFromFormat()

turboxx:
I've added check in datasources when using  \DateTime::createFromFormat($format, $time)
in case of $time doesn't correspond the $format the function returns FALSE

The check should avoid getting 500

and some minor changes

### DIFF
--- a/src/DataSource/ArrayDataSource.php
+++ b/src/DataSource/ArrayDataSource.php
@@ -230,6 +230,10 @@ class ArrayDataSource implements IDataSource
 
 		if ($values['from'] !== NULL && $values['from'] !== '') {
 			$date_from = \DateTime::createFromFormat($format, $values['from']);
+			if (!($date_from instanceof \DateTime)) {
+				/** Date given is not in valid format */
+				return FALSE;
+			}
 			$date_from->setTime(0, 0, 0);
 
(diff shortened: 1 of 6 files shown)


src/DataSource/ArrayDataSource.php
```diff
@@ -220,7 +220,10 @@ public function applyFilterDateRange($row, FilterDateRange $filter)
 
 		if ($values['from'] !== NULL && $values['from'] !== '') {
 			$date_from = \DateTime::createFromFormat($format, $values['from']);
-
+			if (!($date_from instanceof \DateTime)) {
```
paveljanda: Few lines below is a try-catch block, that tries to create `\DateTime` object from other formats. What you added completely discards that feature.

turboxx: Try catch is for $row_value, this check is for eventuality that value in the data source is in different format than $format, which will cause exception on the return bellow

paveljanda: Aah, i see.

juniwalk:
@turboxx Hello, could you fix those conflicts?

PR #261, opened 2016-06-10
Parameter for header_escaping

aldria:
I add parameter for column header escaping.
Default value is escape.


paveljanda:
Thanks!

PR #272, opened 2016-06-17
Html::add() replaced with addHtml() or addText()

dg:
This fixes compatibility with Nette 2.4 and possibly fixes vulnerabilities. 

But I am not sure whether all addText() are used correctly.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -26,7 +26,7 @@
 	},
 	"require": {
 		"nette/application" : "^2.3.0",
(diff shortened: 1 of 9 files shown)


dg:
I used the patch for a while and everything looked ok.

btw for full 2.4 support is needed new release of https://github.com/ublaboo/controls

paveljanda:
Thanks a lot!

PR #286, opened 2016-07-11
FilterText: Add isEntire() for filter entire value in one piece

### LINKED ISSUE #262: FilterText - Hledání dle celého stringu
Šlo by přidat to ``setTextFilter()``, který by určil zda se má hledané slovo dělit dle mezer (aktuální chování) a nebo hledat celý string v kuse?
Případně nějaký super přemínač přímo v hlavní třídě..
Bylo by to super. :)
@paveljanda

RiKap:
I hope that code & naming is good for you.
It close #262 & fix bad variable in [ArrayDataSource.php:157](https://github.com/ublaboo/datagrid/blob/master/src/DataSource/ArrayDataSource.php#L157). On that line is  ``$value`` but it should be ``$word``.

### DIFF
--- a/src/DataSource/ArrayDataSource.php
+++ b/src/DataSource/ArrayDataSource.php
@@ -150,7 +150,7 @@ class ArrayDataSource implements IDataSource
 			$condition = $filter->getCondition();
 
 			foreach ($condition as $column => $value) {
-				if ($filter->isEntire()) {
+				if ($filter->hasSplitWordsSearch() === FALSE) {
 					$words = [$value];
 				} else {
 					$words = explode(' ', $value);
--- a/src/DataSource/ArrayDataSource.php
+++ b/src/DataSource/ArrayDataSource.php
@@ -150,11 +150,16 @@ class ArrayDataSource implements IDataSource
 			$condition = $filter->getCondition();
 
 			foreach ($condition as $column => $value) {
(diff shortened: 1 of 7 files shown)


paveljanda:
Thanks a lot!

But can you please change the name? :D To something like this:

```php

public function setSplitWordsSearch(....)
{
    $this->split_words_search = (bool) .....
}


public function hasSplitWordsSearch()
{
    return $this->split_words_search;
}
```

Thanks..

RiKap:
@paveljanda Ok :)
But "split words search" is actual way of filtering.
So I suggest naming like this:
```
public function setEntireWordSearch(....)
{
    $this->entire_word_search = (bool) .....
}


public function hasEntireWordSearch()
{
    return $this->entire_word_search;
}
```

paveljanda:
It is the way it is filtered by now. So it could be by default TRUE.

```php
private $split_words_search = TRUE;

public function setSplitWordsSearch($split_words_search)
{
    $this->split_words_search = (bool) $split_words_search;
}


public function hasSplitWordsSearch()
{
    return $this->split_words_search;
}
```

Entire word is misleading. Entire sentence maybe, but the phrase does not have to be an actual sentence.

RiKap:
@paveljanda Thx :+1:

PR #298, opened 2016-07-27
Update FilterableDataSource.php

attreid:
Repair issue when Nextras will not filter with condition. For example

        $grid->addColumnText('role', Role)
                ->setFilterSelect($roles)
                ->setCondition(function(ICollection $collection, $value)  {
                    return $collection->findBy(['this->roles->id' => $value]);
                });


paveljanda:
I do not want to create conditions caoused by using different data source in general implementation.

Please use `$collection->getQueryBuilder()`:

```php
$filter->setCondition(function ($collection, $value) {
		$collection->getQueryBuilder()->andWhere("%column = %s", 'name', $value);
	});
```

PR #300, opened 2016-07-27
Added DoctrineCollectionDataSource data-source

juniwalk:
Should be fully working for displaying data.

There is currently problem in Doctrine collection "criterias" so filtering and sorting does not work. I hope it will be fixed in the future.

This is needed if you want to display entities from collection as Array data-source is not sufficient.

### DIFF
--- a/src/DataSource/DoctrineCollectionDataSource.php
+++ b/src/DataSource/DoctrineCollectionDataSource.php
@@ -1,0 +1,261 @@
+<?php
+
+/**
+ * @copyright   Copyright (c) 2015 ublaboo <ublaboo@paveljanda.com>
+ * @author      Martin Procházka <juniwalk@outlook.cz>
+ * @package     Ublaboo
+ */
+
+namespace Ublaboo\DataGrid\DataSource;
+
+use Doctrine\Common\Collections\Collection;
+use Doctrine\Common\Collections\Criteria;
+use Ublaboo\DataGrid\Filter;
+use Ublaboo\DataGrid\Utils\Sorting;
+
+final class DoctrineCollectionDataSource extends FilterableDataSource implements IDataSource
+{
+
+	/**
+	 * @var Collection
+	 */
+	private $data_source;
+
+	/**
+	 * @var string
+	 */
+	private $primary_key;
+
+	/**
+	 * @var Criteria
+	 */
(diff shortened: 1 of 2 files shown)


paveljanda:
How much does that differ from `DoctrineDataSource`? (Asking because this data source can maybe extend the `DoctrineDataSource`?).

juniwalk:
@paveljanda Completely, `DoctrineDataSource` works with `QueryBuilder`, `Collection` is basically just fancy array of entitis which happends to be lazy and fetches those entities from database.

src/DataSource/DoctrineCollectionDataSource.php
```diff
@@ -0,0 +1,272 @@
...
+	 */
+	public function __construct(Collection $collection, $primary_key)
+	{
+		$this->criteria = Criteria::create();
+		$this->data_source = $collection;
+		$this->primary_key = $primary_key;
+	}
+
+
+	protected function initialize()
```
paveljanda: Why are you using that method with return statement, when the return value is not used?

Also, can you please rename the method? I just don't like so much these magic names. `initialize` does not tell much about itself..

paveljanda:
Can you please let me know (in that issue) when the Criteria is ready (fixed) to be used? I will merge it then. 

Thanks a lot for PR!

juniwalk:
@paveljanda Well it is little complicated. Criteria do work, it's just ORM implementation of Collection that does not work (it is not translated to SQL), but when using Collection with array it filters and sorts as it should.

Don't think of this PR as extending support for Doctrine/ORM but as adding support for [Doctrine/Collections](https://github.com/doctrine/collections) namely [ArrayCollection](https://github.com/doctrine/collections/blob/master/lib/Doctrine/Common/Collections/ArrayCollection.php) class.

I've just coded this data-source against interface so it can be used with anything! :)

paveljanda:
Also, can you please change the `@author` doc block of `DoctrineCollectionDataSource ` class? :D

juniwalk:
@paveljanda Satisfied? :)

JakubKontra:
@juniwalk awesome boy! :)

juniwalk:
@JakubKontra Yeah, this should be great addition to datagrid :)

PR #315, opened 2016-08-02
DoctrineDataSource: Added data loaded callback

juniwalk:
We should talk about the name of the method/property for this callback as I am not satisfied with this name.

I've tested this on one of my projects and it already cut 14 queries to just 4 thanks to the [post fetching](https://github.com/Kdyby/Doctrine/blob/master/docs/en/optimizing-query-objects.md) the entities for associations in separate query.

This is how I've used it:
```php
$grid->getDataSource()->setLoadedCallback(function ($orders) {
	$this->orderRepository->createQueryBuilder('o', 'o.id')
		->select('partial o.{id}, v, c')
		->leftJoin('o.vehicles', 'v')
		->leftJoin('o.comments', 'c')
		->where('o IN (:orders)')
		->getQuery()
		->setParameter('orders', $orders)
		->getResult();
});
```

//cc @JakubKontra

### DIFF
--- a/src/DataGrid.php
+++ b/src/DataGrid.php
@@ -540,6 +540,19 @@ class DataGrid extends Nette\Application\UI\Control
 	}
 
 
+	/**
+	 * @return DataSource\IDataSource|NULL
+	 */
+	public function getDataSource()
+	{
+		if (!$this->dataModel) {
+			return NULL;
+		}
+
+		return $this->dataModel->getDataSource();
+	}
+
+
 	/********************************************************************************
 	 *                                  TEMPLATING                                  *
 	 ********************************************************************************/
--- a/src/DataSource/DoctrineDataSource.php
+++ b/src/DataSource/DoctrineDataSource.php
@@ -10,14 +10,22 @@
 namespace Ublaboo\DataGrid\DataSource;
 
 use Doctrine\ORM\QueryBuilder;
-use Ublaboo\DataGrid\Filter;
-use Nette\Utils\Callback;
+use Doctrine\ORM\Tools\Pagination\Paginator;
 use Nette\Utils\Strings;
-use Doctrine;
+use Ublaboo\DataGrid\Filter;
 use Ublaboo\DataGrid\Utils\Sorting;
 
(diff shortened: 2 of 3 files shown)


paveljanda:
You can use native nette events instead of these hardcoded callbacks.

juniwalk:
@paveljanda I'd have to use `Nette\Object` or that new trait which is little overkill for simple data source. You don't need multiple callbacks, you need just one, to execute query after data are loaded to post fetch few more data.

juniwalk:
Besides, this has type hinting, is more compatible with IDEs and such :)

juniwalk:
I've renamed it to `setPostFetchCallback`.

juniwalk:
@paveljanda I've done few more code touches, maybe check them out if there is something you don't like.

juniwalk:
If you have entity with `*toMany` associations and you need them loaded, it is much better to initialize those associations after the parent entities are loaded.

To do this simple, use `onDataLoaded` event of the Doctrine data source to post fetch those associated entities.

```php
use Ublaboo\DataGrid\DataGrid;

class SimplePresenter extends BasePresenter
{
	/** @var Doctrine\ORM\EntityRepository */
	private $orderRepository;

	public function createComponentSimpleGrid(string $name)
	{
		$grid = new DataGrid($this, $name);
		$grid->setDataSource($this->orderRepository->createQueryBuilder('o', 'o.id'));
		$grid->getDataSource()->onDataLoaded[] = function (array $orders) {
			$this->orderRepository->createQueryBuilder('o', 'o.id')
				->select('partial o.{id}, i')
				->leftJoin('o.items', 'i')
				->where('o IN (:orders)')
				->setParameter('orders', $orders)
				->getQuery()->getResult();
		};

		$grid->addColumnText('name', 'Name');
	}
}
```

For more information check [Optimizing QueryObjects](https://github.com/Kdyby/Doctrine/blob/master/docs/en/optimizing-query-objects.md) in Kdyby/Doctrine.

paveljanda:
Thanks!

PR #318, opened 2016-08-09
Outer filters hidden when column hidden

jakubvojacek:
I do not think that outer filters should be displayed when curespondive column is hidden.

### DIFF
--- a/src/templates/datagrid.latte
+++ b/src/templates/datagrid.latte
@@ -26,25 +26,29 @@
 			 *}
 			{if $control->hasOuterFilterRendering()}
 				<div class="row row-filters">
-					{var $i = 0}
-					<div class="col-sm-6" n:foreach="$filters as $f">
+					{foreach $filters as $f}
 						{**
-						 * Each fitler is rendered separately in its own template
+						 * Do not display the filter if the column is hidden
 						 *}
-						{var $filter_block = 'filter-' . $f->getKey()}
-						{var $filter_type_block = 'filtertype-' . $f->getType()}
+						{continueIf !array_key_exists($f->getKey(), $columns)}
+						<div class="col-sm-6">
+							{**
+							 * Each fitler is rendered separately in its own template
+							 *}
+							{var $filter_block = 'filter-' . $f->getKey()}
+							{var $filter_type_block = 'filtertype-' . $f->getType()}
 
-						{ifset #$filter_block}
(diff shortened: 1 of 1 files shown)


dakorpar:
This could crash existing grids, for example I have grids where you have filters defined for something that's not displayed at all, and sometimes user still wants to filter something although they don't need to see it in grid.

jakubvojacek:
Well filters are hidden they're not outer and the column is hidden - I would expect that same behavior for outer filters. However I will add method to this PR, something like `Column::hideFilterWhenColumnHidden(TRUE|FALSE)`

dakorpar:
@jakubvojacek @paveljanda  This broke down existing code in my project, probably in some elses to, Default here should be like previous behaviour, and allowing users to hide it itself, plus It would be nice to have this on whole grid for all columns?

src/templates/datagrid.latte
```diff
@@ -26,25 +26,29 @@
...
-						{var $filter_type_block = 'filtertype-' . $f->getType()}
-
-						{ifset #$filter_block}
-							{include #$filter_block, filter => $f, input => $form['filter'][$f->getKey()], outer => TRUE}
-						{else}
-							{ifset #$filter_type_block}
-								{include #$filter_type_block, filter => $f, input => $form['filter'][$f->getKey()], outer => TRUE}
-								{else}
-									{include $f->getTemplate(), filter => $f, input => $form['filter'][$f->getKey()], outer => TRUE}
+						{continueIf !array_key_exists($f->getKey(), $columns)}
```
juniwalk: Hello, could you use `!isset($columns[$f->getKey()]`? It is faster and should not cause any issues in this case. Also you could possibly get filter key from the foreach above.

juniwalk:
Your idea about adding `Column::hideFilterWhenColumnHidden(TRUE|FALSE)` sounds great! Could you implement it?

dakorpar:
Why not both global and on column itself?

juniwalk:
@dakorpar Well no reason, I just tend to keep things simple. This option is of course possible too.

paveljanda:
@dakorpar Any further intentions to continue with this PR? :)

dakorpar:
@paveljanda that's not my PR :)

paveljanda:
@dakorpar Oh, sorry

@jakubvojacek Any further intentions to continue with this PR? :)

PR #319, opened 2016-08-09
Filter.php::isColumnVisible()

jakubvojacek:
This method can be used in DataSources to determine whether the filter should be used. In our case, we're ignoring given filter when the column is hidden

### DIFF
--- a/src/Filter/Filter.php
+++ b/src/Filter/Filter.php
@@ -142,7 +142,17 @@ abstract class Filter extends Nette\Object
 
 
 	/**
-	 * Get fitler value
+	 * Returns whether corespondive column is displayed or hidden by the user
+	 * @return boolean
+	 */
+	public function isColumnVisible()
+	{
+		return array_key_exists($this->key, $this->grid->getColumns());
+	}
+
+
+	/**
+	 * Get filter value
 	 * @return mixed
 	 */
 	public function getValue()


paveljanda:
Why would you add that method do Filter.php?

jakubvojacek:
Because then you can use it for example here:
https://github.com/ublaboo/datagrid/blob/master/src/DataSource/ApiDataSource.php#L128

`if ($filter->isValueSet() && !$filter->hasConditionCallback() && $filter-> isColumnVisible()) {
    ....
`

paveljanda:
I'm sorry, but I don't like this approach. It go against OOP. Filter should not know about his parent (It does now, but don't take advantage of it). And it definitely should not pass it's parent to some random class using public method.

jakubvojacek:
It does not pass the `$grid` reference outside of Filter class, it returns just a simple boolean, just like `isValueSet`. Filter should know whether its value should be used or not. And in some cases you do not want to use filters that are not visible, that's just confusing. 

How else would you like the DataSource to know whether to use given filter, passing $grid reference into DataSource is way worse I guess.

paveljanda:
I would prefer some setter to the whole `DataGrid` class, that would tell whether to pass all filters to `DataModel` or only the ones that have all columns visible. (The `if` statement would be maybe here: https://github.com/ublaboo/datagrid/blob/master/src/DataGrid.php#L1182)

PR #320, opened 2016-08-10
Add summary callback to set if you want to add row into sum.

Attanon:
Little feature for column summary. If you want to exclude some rows from summary. 

Now you can do it with this code.

```
$grid->setColumnsSummary(['amount', 'id'], function (CashItem $item, $columnName) {
		if ($columnName != "amount") {
			return TRUE;
		}
		return !$item->getDeleted();
	});
```

### DIFF
--- a/src/ColumnsSummary.php
+++ b/src/ColumnsSummary.php
@@ -57,18 +57,18 @@ class ColumnsSummary
 	}
 
 	/**
-	 * Check if can add row sum into sums.. check for every summary column
-	 * @param Row    $row
-	 * @param string $column
+	 * Get value from column.. (with callback or classic getValue from row)
+	 * @param Row    	    $row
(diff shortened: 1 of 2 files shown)


paveljanda:
Thanks a lot for your contribution.

I had somethink a little bit different on my mind. I thought there could be a callback, that will *compute* the total amount. It would give user more flexibility. Do you agree? Do you want to change your PR, or should I do that?

Attanon:
Do you mean something like this?

paveljanda:
@Attanon Yes!

Just for info - what is a CashItem?

Attanon:
@paveljanda I am working on cash register in our client's eshop. And I named one item in the register as CashItem. So CashItem has amount, date and some including info. And I need to show sum of all active items (but in grid i need to show even deleted items)

PR #321, opened 2016-08-11
Column fit content width

misaon:
This PR allow to set `setFitContent` method on column => auto width for column.

### DIFF
--- a/assets/src/datagrid.scss
+++ b/assets/src/datagrid.scss
@@ -272,7 +272,7 @@
 			text-align: right;
 		}
 
-		th.fit-content, td.fit-content {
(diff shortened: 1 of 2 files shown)


paveljanda:
Can you please use TAB instead of spaces? Thanks.

Also, can you make it a bit simpler? Somethink like this:

```php
public function setFitContent()
{
    $this->addAttributes(['class' => 'fit-content']);
}
```

Last thing - can you rename the class from 'fit-content' to 'datagrid-fit-content'? Thanks a lot! :)

misaon:
Jasné, zítra to tam máš :)

paveljanda:
Thanks!

Issue #270, opened 2016-06-16
Group action column dissapers when all rows are disabled


kravcik:
![image](https://cloud.githubusercontent.com/assets/10499130/16116548/e90031d8-33cd-11e6-8382-76f8d58ddb66.png)

Cell should stay "empty" and generate this even when row arent avaible for group action:
`<td class="col-checkbox"></td>`

dakorpar:
I believe it's only wrong that in header row there's still checkbox, group action checkbox shouldn't be displayed if there's no group actions available on rows.

kravcik:
@dakorpar: You are right. Maybe some redraw on filter change should be enough.

paveljanda:
Well, that is a situation. I don't want to redraw the header snippet when filtering (filter input could lost it's focus). What should i do with that?

kravcik:
When testing if group action exist - grid should know if group action was set and its empty or its empty and not was set.

`<th n:if="$hasGroupActionOnRows" rowspan="2" class="col-checkbox">`

silly suggest that in addGroupAction() function add some class var groupActionSet:

`<th n:if="$hasGroupActionOnRows && $groupActionSet" rowspan="2" class="col-checkbox">`

dakorpar:
@kravcik Could you describe how to reproduce the problem?

kravcik:
@dakorpar: Add group action. Set select filter on other column, options A & B. A has allowed group action, B hasn't. If you pick only B in filter - no record has checkbox column.

Btw: Refresh page fix the problem.

dakorpar:
@paveljanda What do you think about adding snippet to header row only
`<tr n:snippet="header-row" n:block="header">`
in datagrid.latte instead of 
`<tr n:block="header">`
and redraw only header row there instead of whole header?

paveljanda:
@dakorpar That would still redraw filter inputs and lose input focus. Right?

dakorpar:
@paveljanda true, I'm using outer filters so I don't have this issue :/
but it shouldn't be hard to put this into two rows, so filters are in seperate row, and then it would be possible to refresh only row with column titles.

paveljanda:
@dakorpar You can not redraw column titles. There are also links for sorting data in table below. And you have to redraw these links..

paveljanda:
The main subject of this issue does not appear anymore, closing ti.

PR #324, opened 2016-08-13
Set filters in seperate rows

dakorpar:
Input fielters should have their own row IMHO.
before: 
before

after: 
after

Also this is first step for successfully resolving issue https://github.com/ublaboo/datagrid/issues/270
but it's standalone feature as well so I'm creating PR only for it.

### DIFF
--- a/src/templates/datagrid.latte
+++ b/src/templates/datagrid.latte
@@ -122,7 +122,9 @@
 							</span>
 						</th>
 					</tr>
-					<tr n:block="header">
+					</thead>
+				<thead n:block="header">
+					<tr n:block="header-column-row">
 						<th n:if="$hasGroupActionOnRows" rowspan="2" class="col-checkbox">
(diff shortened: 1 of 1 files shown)


dakorpar:
I added this commit https://github.com/ublaboo/datagrid/pull/324/commits/831609321c0885896369e26f0d93e4603390f5eb after  although I didn't find usages of header  snippet in serverside code, thought maybe someone else is using it in their own code for some custom thing, with this code now it should be backward compatible.

juniwalk:
@paveljanda What do you think about this?

paveljanda:
@juniwalk If the template is altered correctly - sure, why not.

juniwalk:
@paveljanda Well I'm not sure if it is. You had some concerns about input focus in #270 so you should probably check it first.

dakorpar:
@juniwalk it's not affected by this, since this does not redraw anything, and filters would be redrawed only like they were till now. This would actually make it possible to redraw header row/ filter row seperatelly later, but no code for this now.

juniwalk:
@dakorpar I see, please fix indentation (spaces to tabs).

PR #328, opened 2016-08-15
Adding support nextras traversing expression

attreid:
Adding support nextras traversing expression to all methods except applyFilterText and apply filterMultiSelect


paveljanda:
Why not these two?

I can not test that datasource, so can you add tests, that will test the whole nextras datasource?

attreid:
applyFilterText and filterMultiSelect needs 'OR' in mysql query and Nextras dbalCollection does not support 'OR' in findBy method

I will add tests

paveljanda:
Can you please resolve conflicts with ublaboo master branch? Thanks.

attreid:
Nextras Orm doesn't support sqlite, so is possible to add mysql testing database?

paveljanda:
Ok, i will add mysql database to travis.yml.

attreid:
So, how is it going?

dakorpar:
maybe better to use mockery for unit testing then adding database, tests should be easy to set up and fast...

attreid:
ok, which mockery do you prefer?

dakorpar:
it's up to @paveljanda but I believe Mockery is already there...

paveljanda:
I have merged this PR, but can you please sum up, what happend in all these commits? Thaks!

paveljanda:
Aren't there any BC breaks within NextrasDataSource?

attreid:
There aren't any BC break. I only added support to write column in traversing format, except applyFilterText and applyFilterMultiSelect, because these methods needs 'OR' in mysql query and Nextras dbalCollection does not support 'OR' in findBy method.

paveljanda:
Great, thanks!

attreid:
When do you add this to stable version? I would like to add this repo to my projects instead of my fork

paveljanda:
I will do that right away.

attreid:
Thanks!

PR #331, opened 2016-08-17
Add DataSource unit tests

dakorpar:
Here are DibiDataSource unit tests. See if this satisfy you, I'll probably add more tests here to cover more (date range, range...), now it's almost 1:1 on array datasource tests. Mostly I want to make sure with this that you llike this approach so I don't waste time if maybe you have other idea behind testing this
Also I had to add php.ini so that I can load php pdo_sqlite extension. For running tests you have to specify this php.ini file, I edited travis.yml with this info also.


dakorpar:
I thought little bit and did it like this, I think it's much better now since all datasources are tested on one place, and all of them will have coverage of all tests needed, still if we will need to add some specific test for some DataSource shouldn't be a problem

paveljanda:
I just don't understand, why the array data source is tested in class `BaseDataSourceTest`.

dakorpar:
@paveljanda  tests for all bundles are in BaseDataSourceTest, where do you see specific Array test?
All datasource tests are in BaseDataSource, and all DataSources are tested against same data.

dakorpar:
@paveljanda you see any potential problems with this? IMHO it's better to have it like this then to repeat same code for every test...

paveljanda:
Ah I see, you are inserting data from that property. Maybe in future I will put the data in sqlite file and ArrayDataSource will fetch them to array. Yes, that seams better.

Thaks!

PR #337, opened 2016-08-25
DoctrineCollectionDataSource unit tests

dakorpar:
Forgot on this one, but tests are not going through. I've never used Doctrine so I'm not sure if this is implemented correctly, it should be bu one test is not passing and this needs to be fixed


dakorpar:
ignore this wrong one...

PR #339, opened 2016-08-25
Unnecessary check if class exists

MartkCz:


paveljanda:
Why is it necessary? The namespaces does not have to exist if I am not using them..

MartkCz:
Autoloder is called every time you use class* functions, reflection methods, instanceof and more. Instanceof does not cause error if class name not exists.

paveljanda:
Oh sure, i thought there already isn't the class_exists command. Thanks for PR.

paveljanda:
Can you please rosolve conflicta with master? Thanks.

I remember now why there were these class exiats statements. Because of points in Scrutinizer. :) The performance metter has bigger priority though.

MartkCz:
It's now ok?

paveljanda:
Thanks!

PR #351, opened 2016-09-12
Update bower: nette-forms to version ^2.3 (compatible with 2.4)

attreid:
So I tested editing, adding and filters and it seems to be okay


paveljanda:
Thanks a lot!

PR #355, opened 2016-09-14
fix DataGrid::removeColumn()

janatjak:
After call method DataGrid::removeColumn('test') -> method DataGrid::getColumnsVisibility() throw E_NOTICE: Undefined index.
This PR fix this problem.


paveljanda:
Thanks!

PR #358, opened 2016-09-16
Pagination template: fix show 0 items

CZechBoY:
When no items found then show "Items: 0-0 of 0" text instead of "Items: 1-0 of 0".


juniwalk:
Thanks

PR #365, opened 2016-09-20
Add mysqli extension to tests/php.ini

dakorpar:
Now it's possible to run tests on your dev machine with nette tester,
you just have to create mysql db "datagridTest" and mysql user travis with no password and give travis user permissions to datagridTests.
Although I believe better aproach wouuld be through mock objects....


dakorpar:
this change was merged already with exact filter so I'm closing this

PR #367, opened 2016-09-22
Adding addParameters

### LINKED ISSUE #364: Add external parameters to the Column link and Action
When parameters are passed to the action, there no chance to use external parameters.

For example, we are unable to call action like this *Printer:print(12, self::PAGE_SIZE)* where *self:PAGE_SIZE*  is constant.

attreid:
Adding method ->addParameters(array $parameters) to Action and ColumnLink

### DIFF
--- a/src/Column/Action.php
+++ b/src/Column/Action.php
@@ -62,6 +62,10 @@ class Action extends Column
 	 */
 	protected $attributes = [];
 
+	/**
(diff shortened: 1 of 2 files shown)


genesiscz:
How does one add a parameter callback, which would allow me to work with $item as in other places? I need to add more keys of the particular item

PR #369, opened 2016-09-25
Support for filtering by exact value

dakorpar:
https://github.com/ublaboo/datagrid/pull/242
I did this fix with indentation then had problems with squashing commits, and when I did it I forced push, and pull request got closed, here is now all in one commit...


juniwalk:
@dakorpar Alright, use `dev-master` for now. Thanks.

PR #371, opened 2016-09-26
Abbility to hide filter

dakorpar:
Sometimes I need to filter grid and I don't want this filter to be shown

### DIFF
--- a/src/Filter/Filter.php
+++ b/src/Filter/Filter.php
@@ -74,6 +74,10 @@ abstract class Filter extends Nette\Object
 		['class', 'form-control input-sm']
 	];
 
+	/**
+	 * @var bool
+	 */
+	protected $hidden = FALSE;
 
 	/**
 	 * @param DataGrid     $grid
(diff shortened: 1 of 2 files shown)


juniwalk:
You can filter the data in data source. This way anyone can change filter parameters in the url. What do you think @paveljanda ?

dakorpar:
@juniwalk I know that, but this is more elegant solution, I can't see the way that this would cause something bad, and also if someone wants he can easily hide filters when column is hidden with this one. Plus changing datasource when you have many grids is easier if you do it with this option...

juniwalk:
Sorry I am still not convinced this is the right way. I use data filtering in model all the time without problems (even with many grids).

```php
/**
 * @return Doctrine\ORM\QueryBuilder
 */
protected function createModel()
{
	if (!$user = $this->getUser()) {
		return [];
	}

	return parent::createModel()->where('a.user = :user')
		->setParameter('user', $user);
}
```
This is example of how I use it.

dakorpar:
@juniwalk I'm using it like that as well, and now imagine you want to switch datasource for some reason?
You would have to go through each grid and set it correctly everywhere, while in my way your code would work immidiatelly without need to change every datasource. I'm just thinking what's negative side of implementing that and I don't see it...

juniwalk:
@dakorpar I'll let this on @paveljanda to decide. He knows best how this should work. :)

dakorpar:
@jakubvojacek  This could also solve https://github.com/ublaboo/datagrid/pull/318

paveljanda:
I don't really think that people would use this. And your problem lies somewhere else.

It could lead to user disorientation...

PR #372, opened 2016-09-27
fluent setter for setValue on filter

dakorpar:
Was there some reason why this setter wasn't fluent?


juniwalk:
Probably just overlooked, there is more of these issues.

PR #375, opened 2016-09-27
Datagrid aggreggation functions row

dakorpar:
Every datagrid should have summary rows defined and summary row should be through all results not just on current page. I've started to implement this on datagrid, I'm thinking that current ColumnsSummary is not good solution since it only allow you to get current page summary, and interface is not really extendable, so I'm doing this.
This is now just interface proposal and I just want to see if everybody thinks this is usefull
@paveljanda @juniwalk specially :) If yes I'm continuing with my work.
I really believe every datagrid should have summary row on whole datasource and not only displayed page. This pull request is just so I get approval for task, current interface and suggestions on this at all...
I know it's not a quick task, all datasources will have to be extended as well...


juniwalk:
Related older PR #227

dakorpar:
I hope he changed his mind :) Every datagrid has to have summary through all datasource and not on displayed rows, at least all datagrids i was working on with had that, and summary of just displayed rows is actually in 90% of cases useless...

juniwalk:
@dakorpar Have you changed your mind? I was waiting on response from @paveljanda.

dakorpar:
No, was just trying something out and forgot to restore it...
Wierd, if I do force push Reopen and comment is not possible, and pull request is permanently closed, while deleting branch itself is not an issue...

juniwalk:
@dakorpar That is weird, I can force push normally and GitHub just updates the sha of the commit (and the code ofcourse).

dakorpar:
@paveljanda @juniwalk please I need some response over the weekend for this, if you don't want this, I'll have to either implement some events on which user can then interact, or either have this in my own fork and keep merging constantly...

juniwalk:
@dakorpar I agree with you, but since @paveljanda had different feeling when I wanted to implement this, he should have the final say.

Pavel should be more active this weekend so I hope he will respond.

paveljanda:
Both your (@dakorpar and @juniwalk) columns summaries does not solve the problem of only table-view items summary, right? Or am I missing something?

dakorpar:
@paveljanda true, it could be easily implemented but it would be BC since you have this already, although IMHO this should be done on columns, and not datagrid directly...

dakorpar:
@paveljanda @juniwalk ping

paveljanda:
@dakorpar pong.

It won't be IMHO that easy.. And I don't agree - column summary is just summary and that's it. I don't want to change that.

You are talking about more aggregation functions (sum, avg, ..). Sure, that could be implemented in datagrid (something like `addAggregationFunction()`), right? 

So no BC breaks in that area.

Does that sound reasonable @juniwalk?

juniwalk:
@paveljanda Well I think we need both options, table-view summary and data-wide summary.

If @dakorpar feels he can do it using interfaces, two approaches could be implemented so user has an option which summary he wants.

dakorpar:
@juniwalk It's not an issue I have good background of OOP and this is not an issue to implement, it's mostly issue of time since it's not a quick task, and I don't have much time, so if I'm going to do it it will take some time, on the other hand if you want to do it/have more time, I'll be more then happy that someone else do it :)

dakorpar:
My first gues was to do that as sum only, then when I heard addAggregation function my feel was to add function addAggregation($aggregationFunction),  where aggregationFunction is string, but now if I got you right you wish that a $aggregationFunction is object that implements IAgregationFunction?

juniwalk:
We decided to make it completely separated from current summary implementation with different name. So we should incorproate this into plans.

dakorpar:
This is work in progress, I got it work for most datasources, don't know how to do it for Doctrine, and I hope someone comes with better solution for NetteDatabase, other then that is ok, although on some places I've been using hack approach because this can't be properly implemented if we don't do some more BC, some of which would be really huge and on client end all grid initialization would have to be changed....
This, surelly is not perfect, but I think it can work that way, I'm waiting for opinions because I'm tight with time, had to sleep 3 hours less this night to do this :/
I did not implemented this in views yet since that's like the easiest part
@paveljanda  @juniwalk

juniwalk:
@dakorpar Hey man, sorry about your sleep.

I've been going through your code and I see one potential problem with your future doctrine implementation.

You are trying to solve this in SQL which could bring issues when switching databases. best way would be to stick to DQL (query language doctrine uses).

dakorpar:
I agree and I would really like to avoid raw sql wherever possible, DibiFluent implementation is good one, if I could do something like that in all others I would be more then happy, I never used Doctrine so I really don't know how I would write this, I did some google-ing but didn't really find how to use subquery in from, it's not possible directly, and I assume for someone who used doctrine this is 10minutes or less...
If you take a look I did not implement anything in Doctrine at all, it's hanging...

juniwalk:
@dakorpar Well I work with Doctrine but my knowledge is still somewhat limited. I think that best way would be cloning the QueryBuilder instance and then modifying what you need, mainly select.

```php
$qb = clone $this->data_source;
$qb->select($aggregation_type.'() '.$column'); // You don't need to write 'AS' in SQL nor DQL
$qb->getQuery()->getResult();
```

Subquery should be possible but I've never used it.

dakorpar:
@juniwalk thanks for response, FROM clause should be filtered query ($this->data_source), adding new columns to existing query object won't help me there

juniwalk:
@dakorpar That was not adding, `addSelect` is used for adding :) Something like `$qb->from($this->data_source)` should be possible but like I said, I've never used it.

@fprochazka is probably somehone who could help us with this.

dakorpar:
@juniwalk and that's why I shouldn't implement this :)

dakorpar:
As I see Doctrine does not support this. http://stackoverflow.com/questions/19027518/how-to-perform-subquery-for-from-statement-in-doctrine2/27872713#27872713
Only way to implement this is by getSql and then some string work to edit sql and make it work. I would maybe include event here to put query itself for all datasources... After all I have to know will this be accepted so I finish it that way... @paveljanda @juniwalk

dakorpar:
Hm I actually managed to solve it in preety clean way, it passes tests as well.

dakorpar:
@juniwalk @paveljanda 
this is now complete, I'll wrote a little overview.

- Nextras datasource where every aggregate column is queried in seperate query so if one grid has more aggregate functions defined it won't get all in one query but will query db for every one, other then that I think we're good

- We should probably implement callback where user can define it's own query to get aggregate row, but I think it can go in another PR

- Data is displayed as is, no formatting for numbers so avg function can look wierd, any suggestion on that one? Should I add setFormat function or something?

- All datasources are covered with unit tests and all of them are passing, although I would like to have bigger database (100+ rows) so we can test if I didn't use pagination on some datasource (displaying result only displayed on page and not through all table)

- Every datasource returned different type, result is always correct but sometimes I get it as float 230.00, sometimes string "230" and sometimes integer, that's why in unit tests I'm casting result to integer always, maybe I should always use float and cast all columns in DataModel to float, but this should as well

- on aggregation row I'm only displaying numbers, same as summarry row, but maybe I should also say which aggregation function is used for calculation, any idea on UI for that?

- ApiDatasource is not implemented, I don't know which parameters to set to url there, is this just whatever, and is this datasource mostly just as an example or is actually used for someone...

src/DataSource/DoctrineDataSource.php
```diff
@@ -343,4 +349,27 @@ public function getPlaceholder()
...
+	/**
+	 * get aggregation row
+	 * @return array
+	 */
+	public function getAggregationData()
+	{
+		$this->data_source->resetDQLPart('select');
+		$root_alias = $this->data_source->getRootAliases()[0];
+		foreach ($this->aggregations as $column => $aggregation_type) {
+			$query = $this->data_source->select("$aggregation_type($root_alias.$column) AS $column");
```
juniwalk: You should use `$this->checkAliases($column)` so if there is alias already mentioned, it still works, otherwise good job!

src/DataSource/NextrasDataSource.php
```diff
@@ -294,16 +300,54 @@ public function sort(Sorting $sorting)
...
+	}
+
+	/**
+	 * get aggregation row
+	 * @return array
+	 */
+	public function getAggregationData()
+	{
+		if (count($this->aggregations) == 0) {
+			return;
```
juniwalk: Shouldn't you return empty array?

src/DataSource/ApiDataSource.php
```diff
@@ -198,4 +198,22 @@ public function sort(Sorting $sorting)
...
+		// TODO: Implement addAggregationColumn() method.
+	}
+
+	/**
+	 * get aggregation row
+	 * @return array
+	 */
+	public function getAggregationData()
+	{
+		// TODO: Implement getAggregationData() method.
```
juniwalk: I think you should throw `NotImplementedException` or atleast return empty array.

src/DataModel.php
```diff
@@ -143,4 +144,26 @@ public function filterRow(array $condition)
...
+		/** @var Column $column */
+		foreach ($columns as $column) {
+			if ($column->hasAggregationFunction()) {
+				$hasAggregation = TRUE;
+				$aggregationFunction = $column->getAggregationFunction();
+				$this->data_source->addAggregationColumn($aggregationFunction->getAggregationType(), $aggregationFunction->getColumn());
+			}
+		}
+		if (!$hasAggregation) {
+			return FALSE;
```
juniwalk: How about throwing exception and returing only one type (array)?

dakorpar: We shouldn't throw exception if no aggregations are defined. We just should not display them,
I'll return empty array here instead

dakorpar: Ah, you mean we should return empty array from every datasource, and if datasource doesn't return array then we should throw exception. That can work.

src/DataSource/DoctrineDataSource.php
```diff
@@ -365,10 +365,12 @@ public function addAggregationColumn($aggregation_type, $column)
...
 	public function getAggregationData()
 	{
+		if (count($this->aggregations) === 0) {
+			return [];
+		}
 		$this->data_source->resetDQLPart('select');
-		$root_alias = $this->data_source->getRootAliases()[0];
 		foreach ($this->aggregations as $column => $aggregation_type) {
-			$query = $this->data_source->select("$aggregation_type($root_alias.$column) AS $column");
+			$query = $this->data_source->select("$aggregation_type(" . $this->checkAliases($column) . ") AS $column");
```
juniwalk: Just looking at it now, you should probably use `addSelect` instead and be careful with `... AS $column` as it can contain string looking like this: `p.name`. You should sanitize it.

dakorpar: true, but I should get column key here so I can display it correctly on view in correct td

dakorpar:
@juniwalk I got some wierd errors now, can you take a look at datagrid.php:2725?
I did not do any edits regarding this and not sure how this ends up here?

juniwalk:
@dakorpar Can you post the error here?

src/Column/ColumnAggregationFunction.php
```diff
@@ -0,0 +1,63 @@
...
+	 */
+	protected $aggregation_type;
+
+	public static $aggregation_type_sum = 'sum';
+
+	public static $aggregation_type_avg = 'avg';
+
+	public static $aggregation_type_min = 'min';
+
+	public static $aggregation_type_max = 'max';
```
paveljanda: Why public static property? Why not constants?

dakorpar: I have no idea. I ussually use constants for this... But when on foreign project I always take a look how it's done there, saw public static properties on datagrid.php and without thinking further just got them there as well although these are constants. Will change this ASAP

dakorpar:
@juniwalk @paveljanda so some overview? anything else what needs to be done before merging this or?
I don't want that this PR stays here long because of many changes through many files which can cause a lot of conflicts if not merged in time :)

juniwalk:
@dakorpar Well theres still that `return FALSE` I've reviewed before. Also I want @paveljanda to have final say, this is a huge addition by the looks of it and I want him to say that it is ok.

paveljanda:
@dakorpar I will have time at the weekend, then I will add a review. Is that ok? :)

dakorpar:
@paveljanda more then ok, thank you :)
I just don't want to spent more time on it then I already spent I'm really without free time, so I have to be pushy sometimes :)

dakorpar:
@paveljanda I merged master back in so would be nice if you can go over it this weekend, I assume you couldn't make it last wekend. Just make sure this is preety high in priorities since a lot of code changed and a lot of conflicts possible if not merged on time...

paveljanda:
@dakorpar Thanks you for your time, but a made a bit different approach.

What is different:
1, I want users to write their own implementations of aggregations functions (what you did is in most datasources wrong and would not work).
2, The datasources differes so much that there can not be implementation of all aggregation functions in all datasources.
3, I wanted user to choose whether to use all data or only filtered data or just the data on current page.

dakorpar:
@paveljanda filtered/unfiltered/displayed is not really an issue with this approach as well. I agree that it's not perfect for some datasources and for some of them it wouldn't even work, but only possible approach is then to pass complete query (object)/data to datagrid itself. How do you see it that user implement it's own aggregation functions?  Why do you think my implementatioin is wrong in most datasources? I mean like example why is wrong and what you're trying to achieve....

paveljanda:
@dakorpar 

Some examples:
- The NetteDatabaseTableDataSource implementation is not working when data is filtered and also when the data is paginated. The `Selection::aggregation()` method works differently.
- The DibiFluentDataSource would not work with the version of datagrid you were using with paginated data, because there was `$this->data = $this->data_source->fetchAll($offset, $limit)`, so the limit and offset would not affect the resulted SQL.
And so on. I didn't look at all your implementations.

There is a example implementation of AggregationFunction (`Ublaboo\DataGrid\AggregationFunction\FunctionSum`). It is a `SUM` for `DibiFluentDataSource`. And it is so simple! :)

All you need to do for implementing your own aggregation function is for about 10 lines of code. :)

dakorpar:
@paveljanda it's your decission, thanks for reply :) your interface looks nice, but having one query per aggregation function IMHO is not acceptable, it should be all aggregation functions in one query, because this won't work on any bigger table. Performance issue here is to big... I would propose to you some refactoring where you can add columns and process them all at the end...

dakorpar:
@paveljanda I think you missed my message, could you consider some interface when you can get all aggregation functions in one query? I'm sorry but having one query for every column is not acceptable in basically any website that has some data that's not even big. It will be way to slow...
Just imagine having 10 columns and for each aggregation query 1second...

juniwalk:
@dakorpar Hello, he is on holiday so he will be out for a few days.

paveljanda:
@dakorpar That is a great point, I will think about it.

PR #382, opened 2016-10-03
Set filters in separate rows

dakorpar:
https://github.com/ublaboo/datagrid/pull/324

I did push force to set all as one commit and pull request got closed :/


juniwalk:
@dakorpar Really strange, never happened to me.

Issue #100, opened 2016-02-19
QueryObject as DataSource


Achse:
Currently, doctrine is supported via `\Kdyby\Doctrine\QueryBuilder` -> `Ublaboo\DataGrid\DataSource\DoctrineDataSource`

Using `QueryObject` would provide:
1. better paging thanks to: `doCreateCountQuery`,
2. `postFetch`.

Ad 1. - There is a problem with totalCount calculation and if you want to optimaze your query to avoid aditional selects you have to hack it like:
```php
$this->extraordinaryExpensesRepository->createQueryBuilder('e')
	->innerJoin('e.www', 'w')->addSelect('w')
	->innerJoin('w.ccc', 'c')->addSelect('c')
	->innerJoin('e.ttt', 't')->addSelect('t')
	->leftJoin('e.confirmations', 'conf')->addSelect('conf')
	->getQuery()->getResult();

return $this->extraordinaryExpensesRepository->createQueryBuilder('e');
```

paveljanda:
How much would be "DoctrineQueryObjectDataSource" different implementing from "DoctrineQueryBuilderDataSource"? A do not have any practical experience with Doctrine.

Either you can send a PR, where Doctrine data source will be split into these two data sources (version 2 is on its way anyway) or i will look deep in doctrine a couple days/weeks later i will introduce these split datasources. :)

juniwalk:
QueryObject is just wrapper around QueryBuilder,
I don't see how this is useful. You would have to add methods to add filtering to the QueryObject, you would be basically re-implementing QueryBuilder.

By me, this is :-1:

JakubKontra:
I agree with @juniwalk. 
By me, this is :-1:

paveljanda:
Closing this, I agree with @JakubKontra and @juniwalk .

Achse:
You was little bit rash here. Due to weekend I hadn't time to respond.

`QueryObject` is indeed wrapper around `QueryBuilder`. But beside it it a very good pattern of separation queris from "repositories" it also provides `doCountQuery` and `postFetch` functionality whitch AFAIK you cannot easily achieve with just QueryBuilder.

Here I would talk mainly about the first one.

When you need in `DoctrineDataSource` count of rows for creating pagination you do:
```php
$paginator = new Doctrine\ORM\Tools\Pagination\Paginator($this->getQuery());
$count = count($paginator);
```

[...]

PR #383, opened 2016-10-04
Fix for fetch-joined DoctrineDataSource

neo82nd:
Method getCount correctly uses doctrine's Paginator, but getData returns result directly. It's causing wrong pagination in grid if you provide QueryBuilder with joins - shows less rows than it should be.

This commit change getData method to return result via doctrine's Paginator.

@see http://docs.doctrine-project.org/projects/doctrine-orm/en/latest/tutorials/pagination.html

Problem was mentioned in #100 too.

### DIFF
--- a/src/DataSource/DoctrineDataSource.php
+++ b/src/DataSource/DoctrineDataSource.php
@@ -109,10 +109,13 @@ class DoctrineDataSource extends FilterableDataSource implements IDataSource
 	 */
 	public function getData()
 	{
-		$result = $this->getQuery()->getResult();
-		$this->onDataLoaded($result);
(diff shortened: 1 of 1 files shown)


juniwalk:
Hey, thank you for contributing.
I'd be most happy if DataGrid would be able to work with `\Traversable` so you would just return `Iterator` and be done with it. Maybe in the future. :)

/cc @paveljanda

paveljanda:
@juniwalk Iterating over items of some collection is just a part of all bunch of operations (filtering, sorting, pagination, ...)

juniwalk:
@paveljanda True, I don't know how this works in Array data source for example, but in doctrine, data is reurned after filtering (inside data source) so this should not be a problem.

PR #386, opened 2016-10-06
Escape column identifier in DibiFluentDataSource

dakorpar:
When using SQL keyword as column name, in my case 'key', you'll get SQL syntax error when filtering with that column. This fixes it.


paveljanda:
@dakorpar Travis says your PR s broken.. Can you please chat it out? Thanks!

dakorpar:
@paveljanda that's really wierd, tests are passing and code is working perfectly well
screenshot_3
I see what problem could be, it's different version of dibi, but tests are running on my pc, why?
I can solve it by using old deprecated dibi 2 function escape

PR #387, opened 2016-10-06
add block for outer filters, closes #379

### LINKED ISSUE #379: Outer filter templating
In current version, it is not possible to customize outer filter template with template blocks. For example if someone want to use other layout as col-sm-6 or need to manually render more complicated filter. It would be nice feature to modify this part of template with template blocks.

https://github.com/ublaboo/datagrid/blob/master/src/templates/datagrid.latte#L30

dakorpar:
add block for outer filters, closes #379


juniwalk:
Thanks @dakorpar! :)

PR #404, opened 2016-10-31
Add row condition for item detail

roman-vohnik:
Hi,

this PR allows to set callback as condition to render or not to render item detail for a specific row.

It is called directly on ItemDetail, are you ok with this aproach? Or should I rework it in a way like [existing row conditions](https://github.com/ublaboo/datagrid/blob/v4.4.22/src/DataGrid.php#L2701) - new methods DataGrid::allowItemDetail(callable) and Row::hasItemDetail() ?

I like this way more, but I can be wrong.

### DIFF
--- a/src/Column/ItemDetail.php
+++ b/src/Column/ItemDetail.php
@@ -60,7 +60,7 @@ class ItemDetail
 	/**
 	 * @var callable
 	 */
-	protected $render_condition;
+	protected $render_condition_callback;
 
 
 	/**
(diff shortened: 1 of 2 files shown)


juniwalk:
Looks good to me, please wait for @paveljanda to give you his opinion.

paveljanda:
@roman-vohnik Could you please change the "render-condition" to for example "condition-callback"? Render condition is a bit misleading in my opinion. It does not render enything, it just enables/disables the item detail. Condition callback is also not as good as I want, but i have no idea how to name it better.

Anyone with better name?

roman-vohnik:
"renderable_condition_callback" ?

paveljanda:
@roman-vohnik `render_condition_callback`? :D

paveljanda:
Thanks!

PR #406, opened 2016-11-02
Fixed array source getCount method

jakubvojacek:
The getCount was not returning the correct number of results because it was called after limit() where the size $this->data was changed.


paveljanda:
Thanks!

PR #407, opened 2016-11-02
GetColumn object on filter

dakorpar:
On few ocasions I found this could be usefull. Do not see how it can harm :)

### DIFF
--- a/src/Filter/Filter.php
+++ b/src/Filter/Filter.php
@@ -251,6 +251,7 @@ abstract class Filter extends Nette\Object
 	/**
 	 * @param string $name
 	 * @param mixed $value
+	 * @return $this
 	 */
 	public function addAttribute($name, $value)
 	{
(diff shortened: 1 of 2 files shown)


src/Filter/Filter.php
```diff
@@ -263,13 +264,22 @@ public function addAttribute($name, $value)
 	/**
 	 * @return array
 	 */
-	public function getAttribtues()
+	public function getAttributes()
```
juniwalk: You just fixed this in separate pull request. :)

src/Filter/Filter.php
```diff
@@ -263,13 +264,22 @@ public function addAttribute($name, $value)
...
 	{
 		return $this->attributes;
 	}
 
 
 	/**
+	 * @return \Ublaboo\DataGrid\Column\Column
+	 */
+	public function getColumnObject()
```
juniwalk: Why don't you name it just `getColumn`?

src/Filter/Filter.php
dakorpar: ah yeah, probably created branch from that one, since that is merged shouldn't be any problems with this as well :)

src/Filter/Filter.php
dakorpar: Because getColumn already exists and returns string of column in db

paveljanda:
Thanks for the idea, but i won't merge this.

src/Filter/Filter.php
dakorpar: Yeah I know, that would be kinda not to break the chain :) I found it usefull in my project but I actually thought it's going to be rejected...

Issue #393, opened 2016-10-14
Wrong function name deleteSesssionData


wohral:
I found function deleteSesssionData with triple "s". Maybe it's not wrong, but It makes me little bit confused

juniwalk:
@wohral Yeah that is a typo.

PR #408, opened 2016-11-02
refactor getAttribtues to getAttributes

dakorpar:
I didn't see any usage of deprecated on datagrid, but I used default nette way there...
Maybe we can add in which version this will be removed completelly


juniwalk:
Great job! Thanks!

juniwalk:
I used this method to improve #393 so there is no BC break for next version. Thanks for inspiration.

PR #413, opened 2016-11-05
Basic support for related column names

PingusPepan:
Added basic support for syntax $grid->addColumnType('name', 'Title', ':related_table.column') when using Nette Database

### DIFF
--- a/src/Row.php
+++ b/src/Row.php
@@ -92,6 +92,12 @@ class Row extends Nette\Object
 			return $this->item->{$this->formatDibiRowKey($key)};
 
 		} else if ($this->item instanceof ActiveRow) {
+			if (preg_match("/^:([a-zA-Z0-9_$]*)\.([a-zA-Z0-9_$]*)$/", $key, $matches)) {
+				$relatedTable = $matches[1];
+				$relatedColumn = $matches[2];
+
+				return $this->item->related($relatedTable)->fetch()->{$relatedColumn};
+			}
 			return $this->item->{$key};
 
 		} else if ($this->item instanceof Nette\Database\Row) {


dakorpar:
I believe all features has to be supported for all datasources or it won't get implemented at all,
since this is preety specific for Nette database this should be (IMHO) solved in different approach...

PingusPepan:
@dakorpar The dot in column name is now supported only for doctrine ([here](http://ublaboo.org/datagrid/column#sorting) and [here](https://forum.nette.org/cs/25444-ublaboo-datagrid-mocny-rychly-rozsiritelny-hezky-anglicky-dokumentovany-datagrid#p169436)). That's why I added this for Nette Database. I don't know how related tables are handled in different datasources.

PingusPepan:
Seems to work fine for me. Why not test and merge this if there are no problems and add the feature for all datasources somewhere in the future? I don't see any downsides to this approach other than only users of Nette Database would benefit from this

dakorpar:
Well its question for paveljanda and juniwalk, but IMHO problem then is documentation (seperate for every datasource instead on interface level), and if you switch datasource your grid should work same way, I mean all classes using same interface should return same result, but they can have different implementation. If only doctrine has something different then other datasources then thats also wrong IMHO, just my two cents...

dakorpar:
Well on other hand since theres no new public functions there, that would not work on different datasources there is no much damage, so could be ok...

paveljanda:
@PingusPepan Can you please add some tests? I will merge it then.

paveljanda:
@PingusPepan Ping. are you able to add some tests? There are only one remaining issue and this pull-request to complete v5.0 milestone. We will release a new version (v5.0.0) then.

PingusPepan:
Sorry,
I'm very busy now and in the next few days. Release v5.0.0 and I will
provide the tests when I have time. Hopefully it can be ready for next
version release

PR #425, opened 2016-12-07
Make setOpenInNewTab fluent

dakorpar:
just bumped into this one and feels anoying that in code setOpenInNewTab does not return $this
I also added default to be true, I believe that's how is used through grid ussually...


juniwalk:
Thanks

PR #428, opened 2016-12-08
Don't change collation when searching in DibiFluent

dakorpar:
For me search on this way works much better. I can live with forcing utf8_general_ci
But I don't see a point switching to utf8_bin when some UTF8 character is in search?
Any explanation, or anything else? It's wierd now that search become case sensitive immidiatelly when utf-8 character appear, not to mention completelly different behavior...

### DIFF
--- a/src/DataSource/DibiFluentDataSource.php
+++ b/src/DataSource/DibiFluentDataSource.php
@@ -179,13 +179,8 @@ class DibiFluentDataSource extends FilterableDataSource implements IDataSource
 
 			foreach ($words as $word) {
 				$escaped = $this->data_source->getConnection()->getDriver()->escapeLike($word, 0);
+				$or[] = "$column LIKE $escaped";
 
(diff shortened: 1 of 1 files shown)


ingig:
@dakorpar so what´s the solution to make it case insensitive?

dakorpar:
@ingig we're out of our repository :) This is repo for ublaboo datagrid we're using. Here is pull request for that resolves this, I'll see if we can get solution directly or we will have to add this to out own DataSource class...

ingig:
@dakorpar ok, we'll see what will happen

dakorpar:
@paveljanda this is quick one, can you provide some input here?
IMHO datagrid shouldn't force any collation here, user choosed his html encoding, user choosed his database encoding, IMHO datagrid shouldn't do anything here...

juniwalk:
@dakorpar Hello, @paveljanda is currently away for few days. As I agree with you I will merge this. Pavel can fix it if he has different idea.

Thanks.

PR #430, opened 2016-12-12
Allows to disable translator

jakubvojacek:
Previously it was failing due to `Argument 1 passed to Ublaboo\DataGrid\DataGrid::setTranslator() must implement interface Nette\Localization\ITranslator, null given`


juniwalk:
@jakubvojacek Hello, could you write simple test for setting translator to NULL? I just want to me 100% sure it wont cause any problems.

Thanks!

juniwalk:
ping @jakubvojacek

PR #432, opened 2016-12-12
latte translate macro

teekey99:
Fixes following error using Latte ^2.4:

Invoking filters via $template->translate($vars) is deprecated, use ($vars|translate)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,3 +5,4 @@ vendor/
 # You may choose to ignore a library lock file http://getcomposer.org/doc/02-libraries.md#lock-file
 composer.lock
 coverage.xml
+.idea
--- a/src/Components/DataGridPaginator/templates/data_grid_paginator.latte
+++ b/src/Components/DataGridPaginator/templates/data_grid_paginator.latte
@@ -12,7 +12,7 @@
 	{else}
 		<a class="btn btn-sm btn-default ajax" href="{$link('page!', ['page' => $paginator->page - 1])}" rel="prev">
 	{/if}
-	<i n:block = "icon-arrow-left" class="{$icon_prefix}arrow-left"></i> {_'ublaboo_datagrid.previous'}</a>
+	<i n:block = "icon-arrow-left" class="{$icon_prefix}arrow-left"></i> {='ublaboo_datagrid.previous'|translate}</a>
 
 	{foreach $steps as $step}
 		{if $step == $paginator->page}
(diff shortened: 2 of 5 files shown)


paveljanda:
Does that work also for nette 2.3?

teekey99:
@paveljanda Yes it does. Macro "translate" is part of core macros in version 2.3:
https://github.com/nette/latte/blob/v2.3/src/Latte/Macros/CoreMacros.php#L196

I've also tested it in one of my projects running Nette and Latte 2.3 and it works just fine (using https://github.com/kdyby/translation).

teekey99:
@paveljanda I've missed the translate macro in paginator component template... Sorry for that. I've double-checked whole file tree this time.

juniwalk:
@teekey99 Hello, what's with the equal sign? `{='ublaboo_datagrid.show_all_columns'|translate}`

teekey99:
@juniwalk Without it Latte won't recognize the string inside {} brackets, it will simply output {'ublaboo_datagrid.show_all_columns'|translate} in your template as regular text, not as variable to be parsed.

juniwalk:
@teekey99 Ah I see, I am used to write `{_my.translate.anchor}`. Looks like I need to relearn this :D

teekey99:
@juniwalk That's exactly what this PR is about. Latest Nette and Latte produce deprecated error while using {_something.to.translate} syntax instead of {='something.to.translate'|translate] syntax. I really love using Ublaboo Datagrid but this error breaks down on my latest project.

Heads up: Beware of writing {_something.to.translate} without single quotes like this {_'something.to.translate'}. I've already had issues on some projects while writing translate vars without single quotes. Then we had to re-write it in all templates... Not something you really want to do :D

juniwalk:
@teekey99 Damn, looks like I should pay real attention to what am I writing :D

So you consider this finished? Shall I merge it?

teekey99:
@juniwalk Yes, please :)

juniwalk:
Alright, thanks!

dakorpar:
@juniwalk @teekey99 so you wanna say that' using _ for translations is now deprecated?
In latte documentation it still does not say like that?
https://latte.nette.org/en/macros#toc-localization-expression
And I'm always on newest version did not notice this error in my project and I always use {_'string'} ?

juniwalk:
@dakorpar @teekey99 Yeah I am not getting any error either using `Kdyby/Translation:dev-master`.

teekey99:
@dakorpar @juniwalk Well... it seems to be an issue with `Kdyby/Translation` after all. I never rely on anything but stable dependencies. In this case, it seems the error is fixed on `dev-master` only.

https://github.com/Kdyby/Translation/issues/112

So what now? Using the `translate` macro as I've suggested should result in avoiding any similar issues. I agree using `{_'string'}` is convenient but using `{='string'|translate}` is explicit and self-describing. I suggest using it the same way as `{$var|noescape}` rather than `{!$var}`.

dakorpar:
Basically I don't care how this will be here, with this PR nothing is broken, everything is ok, so I don't see any bad side of it, except that this is really not an issue of datagrid but other package...

teekey99:
@dakorpar Yes, that is correct. I assume I shall explore even closed issues on other packages in my dependencies more carefully next time. I only checked open ones this time and had the issue while using datagrid for the very first time. After that in my own templates as well. As it was not fatal error but a warning only I considered the deprecated warning as valid and the easiest way to get rid of it was using the suggested macro in all templates.

PR #439, opened 2017-01-02
Update datagrid.coffee

xnukes:
Pokud v sablone datagridu stromu vlozim do divu dalsi div, nastane problem se selectorem sortablu. Timto je to fixnuto


dakorpar:
Would it be problem to at least have a commit message on English?

paveljanda:
@xnukes Thanks a lot, I will test it as soon as the gulp on my machine starts working.

PR #444, opened 2017-01-07
Backward compatibility for PHP 5.4

Tomas2D:
Hello, these lines add backward compatibility for PHP 5.4.x. 
It can be useful for someone who is not able to have higher version of PHP on production server.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -30,7 +30,7 @@
 		"nette/forms"       : "^2.3|^2.4",
 		"ublaboo/responses" : "~1.0.0",
 		"ublaboo/controls"  : "~1.0.1",
-		"symfony/property-access": "2.8.15"
+		"symfony/property-access": ">=2.8.15"
 	},
 	"require-dev": {
 		"nette/tester"     : "^1.6.1",
--- a/composer.json
+++ b/composer.json
@@ -30,7 +30,7 @@
 		"nette/forms"       : "^2.3|^2.4",
 		"ublaboo/responses" : "~1.0.0",
 		"ublaboo/controls"  : "~1.0.1",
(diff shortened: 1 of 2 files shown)


juniwalk:
As we have dropped support for PHP 5.5 in version 5.0 I'll leave that to @paveljanda to decide.

Tomas2D:
I understand you that you've dropped support of PHP 5.5. I am personally using latest version of PHP etc, but on my one project I had to use PHP 5.4.

TomasVotruba:
Please do not use at least unsupported PHP versions (those red):

image


http://php.net/supported-versions.php

juniwalk:
I think we should not add support for php 5.4, on the contrary we should remove support for php 5.5 and be reading to cut support for the 5.6 in the near future as it is being phased out.

I am closing this.

TomasVotruba:
[Nette 3.0 is going to be mostly PHP 7.1+](https://forum.nette.org/en/28023-nette-3-0-0-alpha-released#p184590), so I think you can bump this even further so it's fully compatible.

Also [Zenify](https://github.com/Zenify) and [Symplify](https://github.com/Symplify) packages are joining on PHP 7.1

juniwalk:
@TomasVotruba I agree, the sooner we get rid of 5.x line, the better.

TomasVotruba:
@juniwalk I'm with you on this. Let me know if you need PR on PHP 7. I'm quite skilled in it :)

juniwalk:
@TomasVotruba I need to consult with @paveljanda about Nette 3 support (#452) so PHP 7 only release should be among the topics to discuss :)

TomasVotruba:
:+1: It is gonna be next big thing :)

PR #447, opened 2017-01-20
Fixed DoctrineDataSource filtering

foxycode:
When you have filter with native and custom condition, native condition uses postional parameter in `QueryBuilder` and custom condition must use named parameter. For example:

```php
public function createConditionYear(QueryBuilder $queryBuilder, $value)
{
	if ($value) {
		$queryBuilder->andWhere('actuarial.createYear = :year')
			->setParameter('year', $value);
	}
}
```

With another native condition, result in DQL would be something like this:

```sql
SELECT actuarial
FROM Actuarial
WHERE actuarial.createYear = :year AND actuarial.type = ?0
```

If native condition is evaluated after named condition, doctrine rewrites first value, in this case year and second value is never set (error is thrown). When named conditions are evaluated after positional, everything works ok.

Better approach could be generating named aliases instead of positional arguments in native conditions.

### DIFF
--- a/src/DataSource/DoctrineDataSource.php
+++ b/src/DataSource/DoctrineDataSource.php
@@ -11,6 +11,7 @@ namespace Ublaboo\DataGrid\DataSource;
 
 use Doctrine\ORM\QueryBuilder;
 use Doctrine\ORM\Tools\Pagination\Paginator;
+use Nette\Utils\Callback;
 use Nette\Utils\Strings;
 use Ublaboo\DataGrid\Filter;
 use Ublaboo\DataGrid\Utils\DateTimeHelper;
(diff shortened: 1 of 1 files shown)


juniwalk:
@paveljanda What do you think of this?

paveljanda:
@juniwalk Generated aliases are not that big change of code. If it won't cause any BC break and fix that issue, sure, why not.

@foxycode Can doctrine query builder take as named parameter a integer? (` :$p` instead of ` ?$p`)

juniwalk:
@paveljanda I am used to use named parameters everywhere so I did not encounter this issue yet. However we can switch to using generating named parameters in data-source too. We could use pattern for example `:param$p`

PR #451, opened 2017-02-03
Inline edit/add button validation scope

lukascellar:
Inline add/edit should not validate form inputs in other containers.


paveljanda:
Thanks!

PR #457, opened 2017-02-08
Fluent interface improvement

dakorpar:
Seems better this way for me...


paveljanda:
@dakorpar Thanks for PR!

PR #459, opened 2017-02-09
Set default value to filter input if filter value was set

dakorpar:
When some value is set to some filter in code that value should be shown in input as well

### DIFF
--- a/src/Filter/Filter.php
+++ b/src/Filter/Filter.php
@@ -319,6 +319,10 @@ abstract class Filter extends Nette\Object
 			$input->setAttribute($key, $value);
 		}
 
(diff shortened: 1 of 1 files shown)


juniwalk:
@dakorpar Hello, this is not done already? My filters do have default values in them.

dakorpar:
@juniwalk not if you set this bye code with 
`$filter->setValue('value');`
then input field in html did not have value attribute set

paveljanda:
Great idea, I like it!

paveljanda:
Actually there is another method for this purpose: `::setDefaultFilter`

Issue #287, opened 2016-07-14
Combination of inline add and group actions causes a bug


krekoun:
When using  inline add (with required fields), group actions and netteForms.js, after submitting group action an alert message from required field in inline add form appears.

PR #460, opened 2017-02-09
Set validation scope to inline edit and group action submit buttons

jan-stanek:
Fixes issue #287


paveljanda:
@jan-stanek Thanks!

PR #463, opened 2017-02-14
Typo fix assable to assemble

### LINKED ISSUE #462: assableFilters what does assable stand for?
I never heard for that word, google translate also, I found only this
http://www.urbandictionary.com/define.php?term=Assable
and just can't find any reason why this method is called that way?
I believe you wanted to call it assembleFilters maybe?
If so I can deprecate this function and name it correctly, just wished to check with you all before...

dakorpar:
fix #462


paveljanda:
Thanks

PR #476, opened 2017-02-28
Fixed support of dibi 2.x.

radekdostal:
This PR fixes "DataGrid can not take [DibiFluent] as data source" exception, if dibi 2.x is used.

### DIFF
--- a/src/DataModel.php
+++ b/src/DataModel.php
@@ -10,6 +10,7 @@ namespace Ublaboo\DataGrid;
 
 use Nette;
 use Dibi;
+use DibiFluent;
 use Nette\Database\Table\Selection;
 use Doctrine\Common\Collections\Collection;
 use Doctrine\ORM\QueryBuilder;
@@ -61,7 +62,7 @@ final class DataModel extends Nette\Object
 		} else if (is_array($source)) {
 			$source = new DataSource\ArrayDataSource($source);
 
-		} else if ($source instanceof Dibi\Fluent) {
+		} else if ($source instanceof DibiFluent) {
 			$driver = $source->getConnection()->getDriver();
 
 			if ($driver instanceof Dibi\Drivers\OdbcDriver) {


juniwalk:
Wouldn't this broke support for Dibi 3.0? I don't know what @paveljanda had in mind for dibi.

Better way would be to add it as OR to take both options, like that:

```php
} else if (is_array($source)) {
	$source = new DataSource\ArrayDataSource($source);
 
} else if ($source instanceof Dibi\Fluent || $source instanceof DibiFluent) {
	$driver = $source->getConnection()->getDriver();

	if ($driver instanceof Dibi\Drivers\OdbcDriver) {
```

paveljanda:
@juniwalk I would use the `OR`. It is the best options. We sure should support Dibi 3 and 2

radekdostal:
Dibi 3 defines class alias DibiFluent for Dibi\Fluent.

juniwalk:
@radekdostal But we can't be sure it will be there forever.

radekdostal:
@juniwalk It is a question for @dg, if at least dibi 3.x will be compatible with dibi 2.x.

dakorpar:
IMHO if Dibi drops support datagrid should also...
If someone has to use old version of dibi, then he also has to use older version of datagrid...
It's good to support till some time, but I don't see much sense in supporting old obsolete libraries...

paveljanda:
@dakorpar aarkling Dibi 2.x is not obsolete

juniwalk:
@paveljanda Well, [it is obsolete](https://github.com/dg/dibi/releases/tag/v2.3.5) but it is used in many projects :)

juniwalk:
@radekdostal Please, use the `OR` on that alias and it will be merged. However, Dibi 2.x support will be dropped in Datagrid 6.x

radekdostal:
I am sorry for the new PR, but I have deleted the corresponding branch.
See https://github.com/ublaboo/datagrid/pull/480.

PR #479, opened 2017-03-01
Fixed filter escaping for dibi 2.x.

radekdostal:
Class Dibi\Helpers doesn't exists in dibi 2.x.
Related to https://github.com/ublaboo/datagrid/issues/477.


juniwalk:
@radekdostal Please save the driver to `$driver` variable outside the foreach so it is not called each loop.

juniwalk:
@radekdostal Please, include that call on line [188](https://github.com/radekdostal/datagrid/blob/429d166d7c318aa1dbcf7ae0da0d7c27a6b8a799/src/DataSource/DibiFluentDataSource.php#L188). Thanks :)

juniwalk:
Great, thanks!

PR #480, opened 2017-03-01
Fixed support of dibi 2.x.

radekdostal:
This PR fixes "DataGrid can not take [DibiFluent] as data source" exception, if dibi 2.x is used.


juniwalk:
Alright, thanks.

PR #483, opened 2017-03-02
Fix "The variable ‚rows‘ already exists"

temistokles:


paveljanda:
Hm, I know. But it is an issue in nette components. :/

@juniwalk Your opinion?

juniwalk:
@paveljanda Well there is not much of a choice. If there already is any variable this is the only way to overwrite it, as far as I know.

juniwalk:
That `$this->getTemplate()` should be refactored into variable. Soooo much calls for no reason what so ever.

src/DataGrid.php
```diff
@@ -511,34 +511,34 @@ public function render()
...
-		$this->getTemplate()->add('hasGroupActionOnRows', $hasGroupActionOnRows);
+		$this->getTemplate()->hasGroupActionOnRows = $hasGroupActionOnRows;
 
 		/**
 		 * Walkaround for Latte (does not know $form in snippet in {form} etc)
 		 */
-		$this->getTemplate()->add('filter', $this['filter']);
-
+		$this->getTemplate()->filter = $this['filter'];
+
```
paveljanda: Can you please remove this empty line identation? Thanks.

temistokles: There you go :)

radekdostal:
Alternative form of setting is
```php
$this->getTemplate()->setParameters([
  'variable1' => $value1,
  'variable2' => $value2,
  ...
]);
```

paveljanda:
Thanks!

juniwalk:
@radekdostal Does it scream with errors if there already is parameter with the same name?

radekdostal:
@juniwalk I think that add() method throws exception when the variable already exists, but setTemplate() allows it to overwrite.

PR #486, opened 2017-03-06
Fix per-page submit

RiKap:
In newest nette it is submit button, not input ....
Dont forget to regenerate min. version.


paveljanda:
@RiKap Thanks!

PR #490, opened 2017-03-07
Fixed count after data filtering

sallyx:
ArrayDataSource returned the same count for filtered and nofiltered
data.
This commit fixes it by changing count after filtering.

### DIFF
--- a/src/DataSource/ArrayDataSource.php
+++ b/src/DataSource/ArrayDataSource.php
@@ -41,8 +41,7 @@ class ArrayDataSource implements IDataSource
 	 */
 	public function __construct(array $data_source)
 	{
-		$this->data = $data_source;
-		$this->count = sizeof($data_source);
+		$this->setData($data_source);
 	}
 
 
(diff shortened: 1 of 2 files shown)


juniwalk:
Hi, could you fix coding style? Add empty spaces between the count call. Thanks.

sallyx:
You ment ...around the count call. I have done more changes thought.

juniwalk:
@sallyx Sorry I don't like it this way. You call count too many times unnecessarily. You just need to call count on data when the data is being outputted to datagrid.

Check how [DoctrineCollectionDataSource](https://github.com/ublaboo/datagrid/blob/v5.x/src/DataSource/DoctrineCollectionDataSource.php#L68) deals with this.

sallyx:
Done.

juniwalk:
Sorry for being picky, but that private data setter is not really necessary, is it?

sallyx:
Its not necessary, but Its a good place for testing and/or debugging. It would make more sense as protected, but I don't like the idea fixing bug and changing API in one commit. 
So, If you accept this request, the next step would be make it protected.

juniwalk:
@paveljanda What is your point of view on this?

paveljanda:
@sallyx Can you please wrote tests for the array data source, where the previous version is broken and your changes passes? Thanks!

I would remove the `setData` mathod too. It does not bring any enhancement.

sallyx:
The method helped me find out, how many times the data is set compared to getCount(). It was a real enhancement to have this function.
Beside debuging and testing purposes, it also help  with [separation between intention and implementation](https://martinfowler.com/bliki/FunctionLength.html).

sallyx:
Any progress here?

suplere:
A nestačí jen úprava ve funkci public function filter(array $filters), kde před return $this; dát 

**$this->count = sizeof($this->data);**

paveljanda:
Tests added, fix is done. If it is more readable for you with the `setData()` method, why not.

Thanks for the work. :)

PR #492, opened 2017-03-09
Add selected rows count

dakorpar:
Add selected rows count after group actions controls. I don't work with coffee script, only js and ts so someone should probably do something about that and maybe there should be some text like (rows selected), although I like this form as well. In any way I believe this info should be somehow displayed near group action inputs


dakorpar:
@paveljanda any notes?

dakorpar:
@paveljanda @juniwalk any info on this?

paveljanda:
Thanks for the idea, I have implemented in coffee script and enhanced it a little bit. :)

PR #494, opened 2017-03-12
Fix unnecessary setting up template while refreshing state

FVesely:


paveljanda:
@FVesely What should that do? :)

FVesely:
@paveljanda Sorry, maybe I should open an issue first. Without this PR for example an unnecessary select to database is triggered while refreshing state.

paveljanda:
@FVesely Aaah, I see. Shouldn't be that call just after the `$this->onRender($this);`? I mean after the filter assembly.

FVesely:
@paveljanda I don't think so. All required functions are called in handleRefreshState function.

juniwalk:
@FVesely @paveljanda Anything new? How should it be then?

paveljanda:
@FVesely Shouldn't be the `return` in `::render()` method called before `$this->onRender($this);`?

FVesely:
@paveljanda Yes, it should and it's called before `$this->onRender($this);` because it doesn't render anything.

paveljanda:
I like it. Can you maybe call the property `only_refresh_state`?

f3l1x:
I think we can close it, WDYT @paveljanda?

paveljanda:
Yes

Issue #402, opened 2016-10-26
Support hiding columns when $grid->setRememberState(FALSE)


matoni555:
It is possible to support hiding columns while $grid->setRememberState(FALSE) is set?
I use $grid->setRememberState(FALSE) because I don't want to save grid state permanently.

Thank you
Tomáš Matonoha

juniwalk:
@matoni555 Hello,
only way this would be possible is to add information about hidden column into URL. This is something @paveljanda should decide.

dakorpar:
I see this could be useful in url

paveljanda:
There are more looks on that situation (setRememberState(<bool>)). Things, that could maybe stand alone:
1, Use session and remember stuff (in general)
2, Remember filters
3, Remember pagination
4, Remember page
5, Remember per-page
6, Remember hidden columns

And the question is - should they stand alone?

dakorpar:
Well filters and pagination has to it's crucial because you have to be able to copy paste link to someone.
hidden columns could be usefull but it's not crucial...
In setRememberState would also be usefull to define which parts you want to save as well and which maybe not. Also IMHO if something is saved in session as soon as restored url should change to that as well...

PR #499, opened 2017-03-16
Add support for hidden columns without session

marten-cz:
#402 

Tests are not failing. Tested with session enabled and disabled as well. Should be working.

### DIFF
--- a/src/DataGrid.php
+++ b/src/DataGrid.php
@@ -136,6 +136,18 @@ class DataGrid extends Nette\Application\UI\Control
 	public $filter = [];
 
 	/**
+	 * @var array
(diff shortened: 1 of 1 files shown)


juniwalk:
Hello, I like your refactoring but I'd like @paveljanda to check it out first before merging.

paveljanda:
I like the idea.

I don't like one thing though:
- Calling `handleRefreshState` in `::handleHideColumn()`

Is it necessary? Can we make it better? @marten-cz

paveljanda:
This feature won't make it's way into the production code. Sorry but thanks for the idea @marten-cz

PR #500, opened 2017-03-17
Update datagrid.latte

dkocian:
In Firefox that caused a problem with displaying border around <tr n:block="header-column-row">

### DIFF
--- a/src/templates/datagrid.latte
+++ b/src/templates/datagrid.latte
@@ -140,7 +140,7 @@
 						</th>
 					</tr>
 					<tr n:block="header-column-row">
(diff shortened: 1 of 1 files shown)


juniwalk:
Hello, I think that rowspan is there because of inline filters. Did you make sure there is no problem with your code?

dkocian:
Hello, I forgot the inline filters. I've put the rowspan=2 in the condition to appear only when it's active. Always display rowspan=2 is an error. You can check in the Firefox at https://ublaboo.org/datagrid/group-action between the header and the first line missing the margin. In Chrome this works fine in both case.

juniwalk:
You can you [n:attr](https://latte.nette.org/en/macros#toc-n-attr) for conditioned attributes, it should be better.

juniwalk:
Have you tested it? Does it work as expected?:)

dkocian:
Yes. It looks alright.

paveljanda:
Cool, thanks!

PR #501, opened 2017-03-21
Records count is visible only if it is larger than zero. Removed brac…

radekdostal:
…kets around the records count. IMHO it is not necessary to display "items 0 - 0 from 0", if no record was selected.


juniwalk:
I like this more than #515, however I already merged it. I will definitely use this in version 6 but I am thinking of closing this for 5.x and leaving it as is.

What do you think?

radekdostal:
Yes, I agree.

juniwalk:
Alright, thank you anyway :)

PR #502, opened 2017-03-21
Records count is visible only if it is larger than zero. Removed brac...

radekdostal:
…kets around the records count. IMHO it is not necessary to display "items 0 - 0 from 0", if no record was selected.


juniwalk:
Thank you.

PR #504, opened 2017-03-24
Fix datagrid removing original query from URL

temistokles:
Datagrid automatically removes original URL query via history API on refreshState. This leaves the original query present and only adds necessary variables

### DIFF
--- a/assets/dist/datagrid.js
+++ b/assets/dist/datagrid.js
@@ -383,8 +383,20 @@ $.nette.ext('datagrid.url', {
       if (window.history.pushState) {
         host = window.location.protocol + "//" + window.location.host;
         path = window.location.pathname;
-        query = window.datagridSerializeUrl(payload.state).replace(/&+$/gm, '');
-        if (query) {
+     
+ 	var originalQueryParts = window.location.search.replace("?", '').split('&');
+        var urlQuery = {};
+        var i;
+        for (i = 0; i < originalQueryParts.length; ++i) {
+            var item = originalQueryParts[i].split('=');
+            urlQuery[item[0]] = typeof(item[1]) !== 'undefined' ? item[1] : '';
+        }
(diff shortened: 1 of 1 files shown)


stewestw:
Hi pavel,
we have a problem with removing original url params from url. for example we have grid implemented in component which is integrated on link <domain>/admin/core/attributes/edit?id=2
but when we click on inline row edit it set link to <domain>/admin/core/attributes/edit. have you got any suggestions how about fix this please?
thank you

paveljanda:
@stewestw Hello, the problem is, datagrid wants to manipulate the URL, that is obvious. You cat either disable it or set needed parent parameters as persistent.

Did it help?

paveljanda:
I don't know whether it is a good idea to save the original query hence it is not meant to be saved in presenter state.

paveljanda:
@juniwalk What is your opinion?

stewestw:
there will be a problem if i use the param as persistent. please have u got an example how to disable manipulating on this part? thank you for quick response

temistokles:
@paveljanda the patch should fix this for you and you do not need to disable anything

juniwalk:
@paveljanda Well, action and persistent parameters should probably be kept, as other parameters are not expected in php, i would ignore them.

Is this what you wanted to hear?:)

paveljanda:
@juniwalk Yes.

Sorry guys, that is exactly the point of persistent parameters.

@stewestw Take a look on `DataGrid::setRefreshUrl()`, `DataGrid::setRememberState()` methods and play with them a bit.

stewestw:
thank you guys for navigate me. 
the solution which works for me is:
$grid
            ->addInlineEdit()
            ->onControlAdd[] = function($container) use ($grid) {
                $grid -> setRefreshUrl(FALSE);
             /** $container fields */
            };

PR #505, opened 2017-03-24
DoctrineDataSource: use Paginator only when needed

vitkutny:
Paginator is used only if join/groupBy is used in query. For simple data without join it is unnecessarily slow AF.


juniwalk:
Hello, thank you! Could you also send this into v5.x branch as this is for future version that has long way to go?

vitkutny:
ofc, here you go https://github.com/ublaboo/datagrid/pull/506

PR #506, opened 2017-03-24
DoctrineDataSource: use Paginator only when needed

vitkutny:
Paginator is used only if join/groupBy is used in query. For simple data without join it is unnecessarily slow AF.


juniwalk:
Great job! Thank you :)

PR #507, opened 2017-03-27
Add n:block="tbody" to tag tbody

dkocian:
Every parts of table has own block but tbody not. With tbody block can create own table structure.


juniwalk:
Alright, thanks!

radekdostal:
This change probably cause problems with rendering: https://github.com/ublaboo/datagrid/issues/519

PR #510, opened 2017-03-31
Fix group action multiselect not displaying

jan-stanek:
Fixes problem with displaying multiselect in group actions.

Please check if it doesn't break something else.


jan-stanek:
Could you please merge it? It works perfectly for me.

paveljanda:
What was the problem? Can you record an video of your screen where the group action is broken? Thanks a lot. :)

jan-stanek:
Multiselect remains hidden, because it doesn't have any id.
screen

paveljanda:
group_action

Working fine on ublaboo.org..

jan-stanek:
But nowhere else :-D Only difference, I've found, is the missing id. 

I tried removing other js libraries but nothing helped.

paveljanda:
@jan-stanek What do you mean by nowhere else? It works within my projects.. Can you send me a simple sandbox-like app where you can show me that behaviour?

jan-stanek:
I will create something, or you can try it in our public test app: https://srs.skauting.cz/, log in with this username: kraj.vary and password: vary.Web2 and then it is on this page: https://srs.skauting.cz/admin/users/ for "Nastavit role".

jan-stanek:
I've already tried it on a simple page when I was creating this pull request and it was the same. I can't find it now, but I can make a new one next weekend.

paveljanda:
@jan-stanek Yes, please make a new one.. Try to use the latest datagrid version + latest assets. Thanks. :)

jan-stanek:
I found that it stops working after including one of these libraries: smalot/bootstrap-datetimepicker, kartik-v/bootstrap-fileinput
[datagrid-test.zip](https://github.com/ublaboo/datagrid/files/1141624/datagrid-test.zip)

PR #513, opened 2017-04-09
aggregation support for doctrine source

chapcz:


juniwalk:
Thank you

juniwalk:
You could have add querybuilder to default SUM function when I think about it. Will you finish it? :)

See https://github.com/ublaboo/datagrid/blob/v5.x/src/AggregationFunction/FunctionSum.php#L47

chapcz:
Like this?  https://github.com/ublaboo/datagrid/pull/514

PR #514, opened 2017-04-10
Sum support for Doctrine data source

chapcz:
- Added Doctrine QueryBuilder as possible source.
- Optional filter data type (doctrine supported) 
Example usage:
`$grid->addAggregationFunction('price', new FunctionSum('u.price', IAggregationFunction::DATA_TYPE_FILTERED));`
or
`$grid->addAggregationFunction('price', new FunctionSum('u.price', IAggregationFunction::DATA_TYPE_ALL));`

### DIFF
--- a/src/AggregationFunction/FunctionSum.php
+++ b/src/AggregationFunction/FunctionSum.php
@@ -61,8 +61,11 @@ class FunctionSum implements IAggregationFunction
 		}
 		if ($data_source instanceof \Doctrine\ORM\QueryBuilder) {
 			$data_source_clone = clone $data_source;
+			$column = \Nette\Utils\Strings::contains($this->column, '.')
+				? $this->column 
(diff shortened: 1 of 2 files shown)


src/AggregationFunction/FunctionSum.php
```diff
@@ -53,6 +59,16 @@ public function processDataSource($data_source)
 				->fetch()
 				->sum;
 		}
+		if ($data_source instanceof \Doctrine\ORM\QueryBuilder) {
+			$data_source_clone = clone $data_source;
```
juniwalk: I think `$queryBuilder` should be cloned in the DoctrineDataSource itself as it should keep its internal data consistent and thus make sure it does not expose them.

chapcz: When I used only original QB, I had problems with select part - so I only created clone of QB. I sent something which might be better.

PR #521, opened 2017-04-30
Add setPrompt funtion for FilterSelect

kuchy:
I know that something like this work (found after implemented setPrompt...):

`$grid->addFilterSelect('foo','foo',array(''=> '-----','1'=>1,'2'=>2));`

But based on nette form I expected this:

`$grid->addFilterSelect('foo','foo',array('1'=>1,'2'=>2))->setPrompt('-----');`


paveljanda:
Thanks!

PR #526, opened 2017-05-05
fix datagrid-column-header-additions icon position with fitContent method enabled

misaon:


paveljanda:
I like it being on the right side. :)

PR #528, opened 2017-05-09
Typo in doc: replace walkaround with workaround

honzalilak:
See https://github.com/ublaboo/datagrid/pull/527


paveljanda:
Thank you`

PR #532, opened 2017-05-15
Support groupped nette database table data source

temistokles:
Inspired by https://forum.nette.org/cs/28277-zly-count-ublaaboo-datagrid-pri-nettedatabasetabledatasource-s-groupovanim-stlpcu-a-countom-nad-danym-stlpcom, but support of multiple groupped columns added


paveljanda:
Thanks a lot!

PR #538, opened 2017-06-01
Fixed column filter reset

vvoody-:
This extension `$.nette.ext('datagrid.reset-filter-by-column',{..});` is not working properly when Ublaboo\DataGrid\DataGrid component is not in root of component tree. For example in url is `do=grid-grid-resetFilter` and extension is trying to find and replace `do=grid-resetFilter` instead.

In pull request Im adding new payload parameter `_datagrid_full_name` with fully-qualified name of DataGrid component.


paveljanda:
It is working for me (Presenter -> component -> component factory which creates datagrid).

Can you please send me sandbox-like project where that will not work properly? Thanks

vvoody-:
https://github.com/vvoody-/datagrid-bug-report

Bug reproduction steps:
1. filter grid by at least 2 columns, strictly by ajax requests
2. click one of the "x" for reseting single column filter. Whole filter is now reseted. Expected is reset of only one column.

Point is to use only ajax to experience this bug. Full page refresh will generate proper links for column reseting. First ajax request (sort, filter...) will broke filter reset links, but only in case of datagrid not in root of component tree.

peldax:
@paveljanda 

Happens to me too. When the component is not in the root of the component tree (is inside another component), link for reseting filter of one column (The X icon on the right) is invalid.

Example:
my component called `grid` is inside another component `gridContainer`.

Generated URL for this action is 
`do=gridContainer-grid-resetFilter&grid-key=price`
Should be 
`do=gridContainer-grid-resetFilter&gridContainer-grid-key=price`

Current URL is not valid and Nette throws an error, because it tries to pass parameter to the `grid` component.
```
Nette\DI\MissingServiceException
Service 'grid' not found.
```

Easier way to fix this issue is to replace function `getName()` with `lookupPath()` in class Datagrid, row 2526.

peldax:
Alright, other links are generated in template using n:macro, that's the reason why other links work.

peldax:
I created PR with my solution.

#627

f3l1x:
It's resolved via https://github.com/contributte/datagrid/pull/787

PR #542, opened 2017-06-12
Added ability to set the number of columns in the outer filters

proki:


paveljanda:
Great idea!
Can you please add a check (throw an `\InvalidArgumentException`) to the `::setOuterFilterColumnsCount()` method that the given number is one of following numbers: 1, 2, 3, 4, 6, 12?

Thanks

proki:
Added a check

paveljanda:
Supercool

PR #549, opened 2017-06-22
Fix - action link

novotnej:
Grid action is not attempting to link to itself first


fmasa:
@novotnej Do you have any reason not to use composition?

```php
class MyGrid extends Control
{
    public function handleAction() {}

    public function handleAnotherAction() {}

    protected function createComponentGrid(): DataGrid {}
}
```

paveljanda:
@fmasa Showed a solution I like better. Extending DataGrid class is not a great practice.

PR #551, opened 2017-07-06
Send payload as JSON from handleEdit

martinadamec:
Hi, I have problem, when I´m using editable cols. In payload is not property  `_datagrid_editable_new_valuecallback`, but I receive whole HTML page of current view. So I edit this handle and I´m sending payload manually as json.


paveljanda:
It work properly on docu web (https://ublaboo.org/) - what is wrong there?

PR #554, opened 2017-07-19
Update GroupActionCollection.php

vojtys:
na PHP/7.0.18

syntax error, unexpected 'const' (T_CONST), expecting variable (T_VARIABLE)


paveljanda:
Shoot, sorry, forgot that, it should be in next version. Thanks.

Issue #556, opened 2017-07-25
Notice Undefined property: stdClass::$formsStack


DarkPhenix:
This notice occurs when filtering is used in datagrid v5.4.4, Nette 2.4 and autosubmit is turned on.
This issue is similar to #327.
It seems that snippet _gridSnippets_ is not redrawn during reload.

DarkPhenix:
The issue is caused by instante/extended-form-macros component

PR #557, opened 2017-07-25
Add redraw snippet 'gridSnippets' on reload (#556)

DarkPhenix:


paveljanda:
Why is that?

DarkPhenix:
It is fix for issue #556

DarkPhenix:
The issue is caused by instante/extended-form-macros component so this PR is not necessary.

PR #561, opened 2017-08-08
Datagrid.js: fix datagridGroupActionMultiSelect() when not using selectpicker

vitkutny:
When not using selectpicker do not remove id from group action multiselect. Events `loaded.bs.select`/`rendered.bs.select` are never called and thus nette forms won`t toggle multiselect visibility.


paveljanda:
Can you please modify the `datagrid.coffee` file?

vitkutny:
done

paveljanda:
Thanks

PR #562, opened 2017-08-08
Column - added condition callback to allow/disallow inline editing on each row

proki:


src/Column/Column.php
```diff
@@ -588,12 +593,36 @@ public function getEditableCallback()
...
+	public function setEditableOnConditionCallback(callable $editable_condition_callback)
+	{
+		$this->editable_condition_callback = $editable_condition_callback;
+
+		return $this;
+	}
+
+
+	/**
+	 * @return callable
```
paveljanda: `@return callable|null`, hm?

src/Column/Column.php
```diff
@@ -88,6 +88,11 @@
 	protected $editable_callback;
 
 	/**
+	 * @var callable
```
paveljanda: `callable|null`, hm?

src/Column/Column.php
```diff
@@ -588,12 +593,36 @@ public function getEditableCallback()
 
 
 	/**
+	 * Set inline editing just if condition is truthy
+	 * @param callable $editable_condition_callback
+	 * @return $this
```
paveljanda: Pls, `@return static`

proki:
Fixed typehint, but there is error "Nextras\Dbal\ConnectionException: Access denied for user 'travis'@'%' to database 'datagridTest'". I thing, that is not my error ;)

PR #570, opened 2017-08-31
Validation scope for filter and per page inputs

chapcz:


src/DataGrid.php
```diff
@@ -1624,12 +1626,14 @@ public function createComponentFilter()
...
 			->setTranslator(null);
+		$validationscope[] = $form['per_page'];
 
 		if (!$form->isSubmitted()) {
 			$form['per_page']->setValue($this->getPerPage());
 		}
 
-		$form->addSubmit('per_page_submit', 'ublaboo_datagrid.per_page_submit');
+		$form->addSubmit('per_page_submit', 'ublaboo_datagrid.per_page_submit')
+			->setValidationScope($validationscope);
```
jankonas: Hi, you need to flatten `$validationscope` before setting it. You can use `Arrays::flatten` from `Nette\Utils`.

paveljanda:
Thanks a lot for you work. I looked into it and there will have to be actualy more changes needed. For example:
Validation scope can be set only on submit button. All ajax actions and autosubmits are called on the form, not on some button - there is no general submit. So I will have to add a submit button that will be displayed all the time. I also changed the assets, so the autosubmits are calling submit on these buttons. But! I am now stuck on validating the InlineEdit/Add. It does not throw an validation error even with correct validation scope set. :D I will dive into it tomorrow.

Appreciate your work though :)

paveljanda:
Just for info, I pushed this working version into the branch `validation-scope-fixes`.

PR #573, opened 2017-09-14
New feature + Use single quotes when finding option in inline edit 

### LINKED ISSUE #572: When using setEditableInputTypeSelect (option with string causes js error)
Hi,

I have array in editable select callback like 
```php
['0:00' => '0:00'],
['00:30' => '0:30'],
['01:00' => '01:30']
..
```

When the `setEditableInputTypeSelect` fires, everything at server side went good (values are stored to database), but browser says: that Input could not be find cause JS does not include single quotes.

I will submit PR at datagrid.js

FilipLukac:
New feature Export\ExpoArray

This feature allows you to work with exported array (array of filtered / unfiltered results in grid)

- $include_headers indicates that, your array will contain header element at first position of your array
- $filtered - export only filtered rows

Fixes #572 issue

Signed-off-by: Filip Lukac <filip.lukac@gmail.com>


paveljanda:
@FilipLukac `DataGrid::addExportCallback()` is not enough for you? I think it is usable enough.
Can you send a PR with assets fixed only?

paveljanda:
@FilipLukac Merged commit `5be21c1` (`Use single quotes when finding option in inline edit ...`), thank you!

FilipLukac:
@paveljanda is enough, but I thought this would be better, also, thank you for merging this single commit :)

PR #575, opened 2017-09-17
ColumnsSummary: added support for custom renderer

vitkutny:
- BC break? no
- New functionality? yes

Example usage:

```php
$grid->setColumnsSummary(['quantity', 'totalPriceExcVat', 'totalPriceIncVat'])->setRenderer(function ($summary, $column) {
	if ($column === 'quantity') {
		return sprintf('%d ks', $summary);
	}

	return PriceFormatter::format($summary);
})
```

Result:
summary

### DIFF
--- a/src/ColumnsSummary.php
+++ b/src/ColumnsSummary.php
@@ -8,8 +8,10 @@
 
 namespace Ublaboo\DataGrid;
 
-use Nette\Utils\Callback;
 use Ublaboo\DataGrid\Column\ColumnNumber;
+use Ublaboo\DataGrid\Column\Renderer;
+use Ublaboo\DataGrid\Exception\DataGridColumnRendererException;
+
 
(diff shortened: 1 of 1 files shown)


FilipLukac:
@vitkutny could you please provide us documentation ?

src/ColumnsSummary.php
```diff
@@ -94,6 +101,17 @@ public function add(Row $row)
 	 */
 	public function render($key)
 	{
+		/**
+		 * Renderer function may be used
+		 */
+		try {
+			return $this->useRenderer($key);
```
paveljanda: Wouldn't it be a little bit more readable not to catch an exception but checking for instance `$this->renderer instanceof Renderer ` instead?

paveljanda:
Thanks!

PR #576, opened 2017-09-19
DoctrineCollectionDataSource: non strict comparison

vitkutny:
Datagrid does not work properly with data sources using strict comparison in `\Ublaboo\DataGrid\DataSource\IDataSource::filterOne` method.

Strict comparison in DoctrineCollectionDataSource is done at https://github.com/doctrine/collections/blob/master/lib/Doctrine/Common/Collections/Expr/ClosureExpressionVisitor.php#L134.

Other data sources are not using strict comparison.

There is problem when filtering by id from handle methods which is always string and database id in my case is integer.

Is this the right place for fix or numeric string `$id` should be converted to integer in methods `\Ublaboo\DataGrid\DataGrid::handleGetItemDetail` and `\Ublaboo\DataGrid\DataGrid::redrawItem` at lines with `\Ublaboo\DataGrid\DataGrid::$redraw_item` property assignment?

### DIFF
--- a/src/DataSource/DoctrineCollectionDataSource.php
+++ b/src/DataSource/DoctrineCollectionDataSource.php
@@ -15,6 +15,7 @@ use Ublaboo\DataGrid\Filter;
 use Ublaboo\DataGrid\Utils\DateTimeHelper;
 use Ublaboo\DataGrid\Utils\Sorting;
 
+
 final class DoctrineCollectionDataSource extends FilterableDataSource implements IDataSource, IAggregatable
 {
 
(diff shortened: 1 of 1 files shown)


paveljanda:
@vitkutny Could you possibly add some tests please?

f3l1x:
Is it still relevant @vitkutny?

juniwalk:
@f3l1x I did something similar in #975, this may be no longer relevant.

radimvaculik:
Not relevant anymore. Please close @paveljanda

PR #577, opened 2017-09-19
DataGrid: prevent invalid inline edit/add values processing

vitkutny:
`\Ublaboo\DataGrid\DataGrid::filterSucceeded` method is processed in onSubmit of nette form. This event is fired always even values are invalid. Validation is working on frontend side (if implemented), but it needs to be validated while submitting form as well.


paveljanda:
@vitkutny Thanks

Issue #496, opened 2017-03-14
Text filter doesn't work with DibiFluentDataSource and PostgreSQL


elring:
When searching in text fields, grid in DataSource/DibiFluentDataSource.php in function applyFilterText at line 165 escapes column name with quotation marks. It results in SQL condition ("column" like '%aaa%'). 
But Dibi translates this to ('column' LIKE '%aaa%'), which (at least for PostgreSQL) means comparison of two strings, which results in selection of all records (when second string contains the first one) or empty result otherwise.
When I remove the quotation, everything works as expected. Don't know whether this is error of datagrid or Dibi.

paveljanda:
@elring I would create another data source class just for Postgre that would inherit from normal dibi data source, but would implement it's own `escapeLikeString()` (parent class - `DibiDataSource` would implement it too) and both datasources would be using the method.

Can you send a PR? :)

elring:
OK, I have done this for myself (probably not mergeable I think, but I can make PR if you think it makes sense). 

Made custom DibiFluentPostgreDataSource with only one method applyFilterText where I've commented-out the escape of column name.

Reason for this is, that the escape works correctly (the dibi Postgre Driver escapes the column name into "column"), but later the dibi where condition (in $this->data_source->where($or);) replaces the quotation marks (") with apostrophe ('). So I think this is bug in dibi... If I make simple dibi test with dibi:test('select column from blabla where "column" = 1') it has this bug too.

elring:
Sorry for multi-comment.

This is not bug of Dibi, but a feature (and probably bug in Datagrid). According to source (and doc) of dibi, it doesn't use database-native quotation, but its own style of identifier and character quotation (\`identifier\`, [identifier], 'string', "string"). So Datagrid should use this quotation and not try to use dibi-driver-specific-magic quotation (it is coincidence that it works for Mysql etc. because it use same quotation, but Postgresql use different marks than dibi). 

So correct way is to modify DibiFluentDataSource and always quote with [] or \`\`.

theo1024:
@paveljanda Is there any chance to merge this PR? I will really appreciate that, because I solving the same problem right now.

@elring Great job! Thank you for that!

PR #580, opened 2017-10-04
Fix for non-functional text filter in PostgreSQL + dibi

elring:
Fixes issue #496 + noted problem in Datagrid thread on Nette forum.

Created suggested DibiFluentPostgreDataSource with few modifications:
* avoid db specific escape of column name - use dibi quotation
* use ILIKE for fulltext search (PostgreSQL specific for case insensitive search)


paveljanda:
@elring Thanks!

PR #582, opened 2017-10-13
Fix click tracy bug

kravcik:
Any click on tracy - populate this problem. This commit fix it.

image

```
Uncaught TypeError: Cannot read property 'parentNode' of null
    at getEventDomPath (jsloader-f5339029bb94.js?1507884840:1246)
    at HTMLDocument.<anonymous> (jsloader-f5339029bb94.js?1507884840:1246)
```


paveljanda:
Already fidex

PR #588, opened 2017-11-21
disable validations when changing per page

zemekoule:
When new row or edit row forms are visible, changing per-page setting currently triggers their validation. This in turn prevents changing per-page setting.

My modification fixes it by specifically setting a validation scope on the per-page submit button.

P.S. We spoke about this when we met on Nette Camp this summer.


paveljanda:
Thanks

PR #590, opened 2017-11-23
Fix DibiFluentDataSource like issue

ne0-cz:
https://github.com/ublaboo/datagrid/issues/589

The problem was double escaping and it's only reasonable to leave all escaping to dibi (and not input escaped queries to it). I didn't fix this for PostgreSQL or MS SQL variants since I have no experience with them.


paveljanda:
Thanks!

PR #597, opened 2017-12-15
Add option to set render condition callback on action

jankonas:
Just like you can set render condition for item detail action, you should be able to set render condition for other actions.


paveljanda:
@jankonas Thank you for the idea, I like it. Just a small note aside: Do you think you can put some functionality to a trait so the code won't be repeated across multiple classes? :)

jankonas:
Sure, I'll try to find some time in this week to do this.

jankonas:
I updated the PR with added trait. Sorry for the delay :-)

paveljanda:
@jankonas Awesome. Thanks, man.

PR #598, opened 2017-12-19
Fix: SmartObject is in nette/utils 2.4.0

Spilky:
SmartObject is in nette/utils 2.4.0 (https://github.com/nette/utils/releases/tag/v2.4.0). It is not in version 2.3.10 so that require is not enough.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -27,7 +27,7 @@
 	"require": {
 		"php": "^5.6|^7.0",
 		"nette/application": "^2.3",
-		"nette/utils": ">=2.3.10",
+		"nette/utils": ">=2.4.0",
 		"nette/forms": "^2.3",
 		"ublaboo/responses": "~1.0.0",
 		"symfony/property-access": "~3.0"


paveljanda:
@Spilky Uups, thank you!

Spilky:
@paveljanda I noticed that you fixed this in release 5.4.10, but in release 5.4.9 it is still not working, because you require nette/utils 2.3.10 and it is not enough. Can you fix it in version 5.4.9? Otherwise it will throw exceptions because of using SmartObject which is not in nette/utils until version 2.4.0. Thanks a lot.

juniwalk:
Hello, he cannot fix it back in 5.4.9. You have to upgrade to newer version. That is why there are releases in the first place.

Spilky:
According to this https://stackoverflow.com/questions/24849362/change-connected-commit-on-release-github he can delete that tag and add it to right commit. It is only one commit after so there should be not problem. This is right commit for tag 5.4.9: https://github.com/ublaboo/datagrid/commit/fbd65401a54b66952395982de727b816240a8b5c.

I can not upgrade to newer version because I con not use nette/utils 2.4.0 which is required.

paveljanda:
@Spilky It's moved. You're right, I didn't looked closely into the commit history. :)

PR #604, opened 2018-01-09
Fix if node is null – break while

tomaskavalek:
Fix for "Uncaught TypeError: Cannot read property 'parentNode' of null"


paveljanda:
@tomaskavalek Can you please edit also datagrid.coffee? I will "gulp" the rest. :)

tomaskavalek:
@paveljanda Sorry, my bad 👎 I did it in two separate commits – also not good, I know.

paveljanda:
@tomaskavalek Thanks, it's ok. :) GitHub can squash these commits automatically when merging pull requests.

PR #613, opened 2018-02-22
Dibi 2.x and 3.x compatibility fix

elcheco:
I am using the latest dibi and there were two exceptions thrown in Row.php and DibiFluentDataSource.php files because of version with namespaces or old one without namespaces.

### DIFF
--- a/src/DataSource/DibiFluentDataSource.php
+++ b/src/DataSource/DibiFluentDataSource.php
@@ -35,10 +35,10 @@ class DibiFluentDataSource extends FilterableDataSource implements IDataSource, 
 
 
 	/**
-	 * @param DibiFluent $data_source
+	 * @param DibiFluent|Dibi\Fluent $data_source
 	 * @param string $primary_key
 	 */
-	public function __construct(DibiFluent $data_source, $primary_key)
+	public function __construct($data_source, $primary_key)
 	{
 		$this->data_source = $data_source;
 		$this->primary_key = $primary_key;
--- a/src/Row.php
+++ b/src/Row.php
@@ -17,6 +17,7 @@ use Nette\Utils\Html;
 use Nextras;
 use Ublaboo\DataGrid\Exception\DataGridException;
 use Ublaboo\DataGrid\Utils\PropertyAccessHelper;
+use Dibi\Row as DRow;
 
 class Row
 {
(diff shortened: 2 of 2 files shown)


paveljanda:
It should work fine with class aliases defined in https://github.com/dg/dibi/blob/master/src/loader.php

mkoula:
I have the latest dibi "dev-master", but still I am getting the exception:

**TypeError**

Argument 1 passed to Ublaboo\DataGrid\DataSource\DibiFluentDataSource::__construct() must be an instance of DibiFluent, instance of Dibi\Fluent given, called in /www/shop/vendor/ublaboo/datagrid/src/DataModel.php on line 77

Do I need to load it somehow? it's part of the loader in dibi so I assumed it should be done automatically, but obviously not...

paveljanda:
It's in dibis composer.json file.

mkoula:
The thing is that it should be loaded automatically if I have dibi installed through composer which I have, but I am still getting these exceptions mentioned above...

BigOHenry:
With dibi 3.1 its ok, but with dibi 3.2 getting error:
`Argument 1 passed to Ublaboo\DataGrid\DataSource\DibiFluentDataSource::__construct() must be an instance of DibiFluent, instance of Dibi\Fluent given, called in /var/www/flexii/app/customizations/EvosoftDataSource.php on line 29`

src/DataSource/DibiFluentDataSource.php
```diff
@@ -35,10 +35,10 @@ class DibiFluentDataSource extends FilterableDataSource implements IDataSource,
 
 
 	/**
-	 * @param DibiFluent $data_source
+	 * @param DibiFluent|Dibi\Fluent $data_source
 	 * @param string $primary_key
 	 */
-	public function __construct(DibiFluent $data_source, $primary_key)
+	public function __construct($data_source, $primary_key)
```
paveljanda: Can you please add an argument type check and in case of wrong argument given throw a `\InvalidArgumentException`? Thanks.

f3l1x:
I think it's already fixed in 5.x. Right?

BigOHenry:
Hello, nope. its not still fixed. 
```
User Deprecated
Class DibiFluent  has been renamed to Dibi\Fluent.
```

I have to create my own datasource to work with Dibi\Fluent. Its copy of original one with fixed DibiFluent references...

f3l1x:
Since v6.0 of the datagrid will be datasource for dibi in separate repository. 

Please track this repo: https://github.com/contributte/datagrid-dibi-data-source

PR #628, opened 2018-03-17
fix render url for signal in parent control in production mode

Ivorius:
Viz. https://forum.nette.org/cs/25444-ublaboo-datagrid-mocny-rychly-rozsiritelny-hezky-anglicky-dokumentovany-datagrid?p=26#p195751


paveljanda:
Thanks

Issue #601, opened 2018-01-04
DefaultSort does not work


JohnyRicio:
Hi,

when I use setDefaultSort on init grid, all is ok. But when I set order by column and after reset this, default sort is ignored and data is ordered as was returned from database. I use for datasource kdyby/Dotrine.

Thank you for fix

paveljanda:
It should be fixed by now (https://github.com/ublaboo/datagrid/pull/545)

JohnyRicio:
Fix is not working. Data are always sorted ASC, but I have $grid->setDefaultSort(['id' => 'DESC']);

paveljanda:
@JohnyRicio What version did you try? It is not released yet (just sitting in branch `v5.x`).

JohnyRicio:
Current version what I have is 5.4.13. Yesterday, I updated to this version and I did not check if this fix was be released. Maybe It been fixed. Sorry

paveljanda:
@JohnyRicio I wrote that is has not been released yet (it will be in following days). But you can test the `v5.x` branch (`HEAD`).

JohnyRicio:
@paveljanda Currently I have v5.x-dev 80e5b81 and still not working. Default shorted data is ordered ASC, but I seted it to DESC.

paveljanda:
@JohnyRicio I can't find a commit `80e5b81`. :D Where did you got it?

Can you please create a sandbox-like project with some basic functionality that will show that issue and upload it to github / here with zip? Thanks a lot!

ne0-cz:
I can confirm this issue for 5.5.5.

```
$grid->setPrimaryKey('post_id');
$grid->setDataSource($this->postFacade->getFluent());
$grid->setDefaultSort(['date' => 'DESC']);

$grid->addColumnDateTime('date', 'Date')
            ->setFitContent()
            ->setSortable();
```

5.4.13 acts as expected:
On load: `ORDER BY date DESC`
Click: `ORDER BY post_id`
Click: `ORDER BY date ASC`
Click: `ORDER BY date DESC`
Click: `ORDER BY post_id`
Click: `ORDER BY date ASC`
Click: `ORDER BY date DESC`
[...]

5.5.5 is wrong:
On load: `ORDER BY date DESC`
Click: `ORDER BY date ASC`
Click: `ORDER BY post_id`
Click: `ORDER BY date ASC`
Click: `ORDER BY post_id`
[...]

[...]

PR #630, opened 2018-03-20
reset-sort-when-reseting-filter

proclarush-teonas:
this should repair bug from this comment: https://github.com/ublaboo/datagrid/issues/601#issuecomment-368484053
`It does not work (means the grid default sort is not applied) when reseting filters with the "reset filters" button.`
-> sort will be reset when `reset filter` button is hit


paveljanda:
#601 already closed via #679

PR #634, opened 2018-03-22
Allows set custom paginator template before datagrid component is attached to presenter component

mhucik:
I cant set custom template for paginator component before datagrid is attached to presenter.
This code allows that, so you can make common factory for specific grid design and dont have to set template after this factory is attached to presenter component.


paveljanda:
👍

PR #636, opened 2018-03-23
DibiFluent && Dibi\Fluent support

cniry:


src/AggregationFunction/FunctionSum.php
```diff
@@ -63,7 +62,7 @@ public function getFilterDataType()
 	 */
 	public function processDataSource($dataSource)
 	{
-		if ($dataSource instanceof DibiFluent) {
+		if ($dataSource instanceof Dibi\Fluent || $dataSource instanceof DibiFluent) {
```
paveljanda: Will that work when there is no `use DibiFluent;` statement in the file header?

paveljanda: @cniry ping

elring: I think it is at least missing "use Nette" OR the test should be "$dataSource instanceof \Dibi\Fluent ..." (without any of these changes it doesn't work for me).

f3l1x:
Duplicate? https://github.com/contributte/datagrid/pull/662

f3l1x:
It needs to be rebased, @cniry.

paveljanda:
@cniry ping

PR #637, opened 2018-04-04
 Added noescape macro for columnsSummary

revoke:
Because for rendering I am using setRenderer() callback with my own HTML format


paveljanda:
@revoke Could that be decided by some setter? `ColumnsSummary->setTemplateExcaping($escape = true) `

revoke:
I think it's not necessary because it is a bug fix.

paveljanda:
It is a possible security issue now. I would give it a thumb up when there is a setter for it.

PR #639, opened 2018-04-06
Fixed CSS classes for buttons

radekdostal:
Renamed CSS class btn-xs to btn-sm, removed class btn-default, which doesn't exist in BS4.


paveljanda:
Bootstrap 4 support is ready in branch `bootstrap-4`. Btw, datagrid has to support bootstrap 3 same as 4..

radekdostal:
OK. So add btn-sm to everywhere where the btn-xs class is used? BS4 [dropped btn-xs](http://getbootstrap.com/docs/4.0/migration/#buttons). If both classes exists, BS3 uses btn-xs.

radekdostal:
I will send a new PR.

PR #651, opened 2018-04-25
Empty values are not sending

MilanPala:
When using PHP 7.1 types empty arguments throw an error due to bad type conversion from empty string to int. They must be not sent

`public function handleSort(?int $item_id = NULL, ?int $prev_id = NULL, ?int $next_id = NULL)`


paveljanda:
👍

PR #657, opened 2018-05-18
Toolbar buttons in treeview

peldax:
Hi, this simple PR adds toolbar buttons to tree view template.

I am wondering why isnt treeview also in rendered as table but uses div layout - which is ofc incompatible with table. It completely blocks reusing blocks from parent tempalate - such as odering and filtering - it is all working (for NTDB) when included from parent template, bud disabled for treeview. 

I would create PR for ordering and filtering also, but I dont have much time to spare ATM.


paveljanda:
👍

PR #659, opened 2018-05-22
Fixed return annotation

radekprokes:


paveljanda:
@radekprokes This method return only `static` (`return $this`)

radekprokes:
I know, so I fixed it so and removed the bad callable annotation

f3l1x:
I know it's not part of this PR, but could you please drop PHP 7.2 in Travis? There's unrelated misusage in nextras/orm which failing on PHP 7.2.

paveljanda:
@radekprokes Oh I see, I was confused by the diff colors..
@f3l1x Let's make another PR for that.. https://github.com/contributte/datagrid/pull/660

Issue #661, opened 2018-05-23
Error when sorting with DibiFluent datasource with Dibi 3


elring:
With DibiFluentDataSource I sometimes got error "Class DibiFluent  has been renamed to Dibi\Fluent.".
Error is in method sort in DibiFluentDataSource.php, where it still uses now non-existent class DibiFluent.

PR #662, opened 2018-05-23
Fix for DibiFluent vs Dibi\Fluent sort

elring:
Fix for #661


f3l1x:
👍

paveljanda:
@elring Can you send a PR also to the master branch? :) Thanks!

elring:
@paveljanda It looks to me that master branch doesn't work with old Dibi - only with new Dibi\Fluent variant. So no need for backward compatibility. 

And the replace to Dibi\Fluent was already done - even in the sort function - by you :)
`$reflection = new \ReflectionClass(Fluent::class);`

paveljanda:
@elring Good point.

f3l1x:
Merged. Thanks.

PR #672, opened 2018-06-22
Fix conditionCallback invocation when using DoctrineCollectionDataSource

ondrejmastik:
Now, when using DoctrineCollectionDataSource, application fails on condition callback invocation. it's caused by DoctrineCollectionDataSource having private access to data_source property instead of protected. 

Please merge

Thanks.

### DIFF
--- a/src/DataSource/DoctrineCollectionDataSource.php
+++ b/src/DataSource/DoctrineCollectionDataSource.php
@@ -21,17 +21,17 @@ final class DoctrineCollectionDataSource extends FilterableDataSource implements
 	/**
 	 * @var Collection
 	 */
-	private $data_source;
+	protected $data_source;
 
 	/**
 	 * @var string
 	 */
-	private $primary_key;
+    protected $primary_key;
 
 	/**
 	 * @var Criteria
 	 */
-	private $criteria;
+    protected $criteria;
 
 
 	/**


Naxmars:
Would it be a bad idea to add abstract method getDataSource to FilterableDataSource? Obviously every DataSource should have accessable dataSource for purpose of filter condition callback invocation, so noone would make this mistake again when creating new DataSource

src/DataSource/DoctrineCollectionDataSource.php
```diff
@@ -21,17 +21,17 @@
...
 	 * @var Collection
 	 */
-	private $data_source;
+	protected $data_source;
 
 	/**
 	 * @var string
 	 */
-	private $primary_key;
+    protected $primary_key;
```
paveljanda: Can you convert the identation to tabs?

src/DataSource/DoctrineCollectionDataSource.php
```diff
@@ -21,17 +21,17 @@
...
 
 	/**
 	 * @var Criteria
 	 */
-	private $criteria;
+    protected $criteria;
```
paveljanda: Can you convert the identation to tabs?

ondrejmastik: Aren't they already indented by tabs?

ondrejmastik: Hmm they, arent. Sorry

paveljanda:
👍 after fixing the indentation.

f3l1x:
I've merged it by my own. Thanks anyway @ondrejmastik.

PR #677, opened 2018-06-30
Fix ItemDetailForm for PHP7.2

oldamalec:
[Parent function](https://github.com/nette/component-model/blob/master/src/ComponentModel/Container.php#L105) requires this second parameter.


paveljanda:
👍

PR #679, opened 2018-07-07
Fix default sort when column sort was changed and remember state is enabled.

### LINKED ISSUE #601: DefaultSort does not work

pavelkovar:
Fix #601.


paveljanda:
Looks OK. 👍

PR #682, opened 2018-07-09
ColumnsSummary ->setPositionTop

vladimirbiro:


src/ColumnsSummary.php
```diff
@@ -41,6 +41,11 @@ class ColumnsSummary
 	 */
 	protected $renderer;
 
+	/**
+	 * @var bool
+	 */
+	protected $positionTop;
```
paveljanda: Is there any default value of this property? :)

```php
protected $positionTop = false;
```

vladimirbiro: image

false = false
(other) = true

true:
image

false:
image

paveljanda: There is no initial call of `::setPositionTop()` so the default value is `null`. Please set the default value to false as I mentioned in the first comment.

PR #683, opened 2018-07-10
Add redraw main snippetArea after reload

martinknor:
because i need globalFormStack at footer


f3l1x:
WDYT @paveljanda?

paveljanda:
I think that would redraw filter inputs, wouldn't it?

paveljanda:
It definitely would. I dont think this is a good idea.

f3l1x:
So I'm closing it now. Thank you @martinknor anyway.

PR #689, opened 2018-07-16
Allow zero value in custom editable value

lukascellar:
If html attribute data-datagrid-editable-value="0" then value was loaded from table cell, not data attribute. Value should be loaded from data-datagrid-editable-value attribute if exists.


f3l1x:
👍 

@paveljanda ?

paveljanda:
👍

PR #699, opened 2018-09-05
Bootstrap 4

f3l1x:

PR #700, opened 2018-09-08
Bootstrap 4 (rebased)

f3l1x:
Rebased #699 without merge commits and conflicts. 

BC break: dropped *.min.js and *.min.css


paveljanda:
@f3l1x Why that BC break?

f3l1x:
It's hard to maintain coffee -> JS -> minify, same thing for SASS -> CSS -> minify without propel scripts or sth. How have you did until now?

paveljanda:
Using one old gulpfile..

- Can't github CI do something like that? -> Starting docker-compose based configuration and build a minified assetets inside it?

Hm, it really should not be a part of the vendor. Ok, good.

PR #706, opened 2018-10-09
Fix filter calendar icon

radekdostal:
Fixed missing calendar icon if Bootstrap 4 is used. Bootstrap 3 compatible.


radekdostal:
Would not be better to release new version 5.7.1? This is a screenshot from Chrome and Opera before this fix. FF, IE and Edge are OK.
ublaboo-filter-calendar

PR #712, opened 2018-10-23
fix group text action collision with another datagrid

Matousek:
When you have 2 datagrid on one page. You can find this problem. Problem is with addGroupTextAction. Te hidden text input has same class as all datagrids. When you use group action on 2nd datagrid and 1st datagrid has addGroupTextAction too, text input is opens on 1st datagrid. This fix solves it using  unique classes. (2nd way is edit JS)

### DIFF
--- a/src/GroupAction/GroupActionCollection.php
+++ b/src/GroupAction/GroupActionCollection.php
@@ -19,7 +19,7 @@ class GroupActionCollection
 
 	use SmartObject;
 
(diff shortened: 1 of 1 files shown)


f3l1x:
I think it's already fix, right @paveljanda?

paveljanda:
I am not sure about it. I will test it.

paveljanda:
It's broken still in v5.x and also in master. @Matousek thanks a lot!

PR #714, opened 2018-10-23
Fix Exception throw if parent is null

mlazovla:


f3l1x:
Hi, could you please change target to v5.x instead of master? Thanks.

mlazovla:
Changed from master to v5.x.

f3l1x:
Yep, but you also have to rebase it. :-)

PR #720, opened 2018-10-29
Update src/Traits/TLink.php

MartinSzollos2016:
Fix render url for presenter signal


MartinSzollos2016:
This https://github.com/contributte/datagrid/pull/628 doesn't solve presenter links.

f3l1x:
Could you please add some test? I would be very useful.

paveljanda:
Thanks for the tests. Looks good.
👍

PR #721, opened 2018-10-29
Fix throwing exception on null parent.

mlazovla:
I removed patch on master branch and created new patch (with same changes) to v5.x branch.


f3l1x:
Thanks.

Issue #723, opened 2018-11-07
Bug in DataGridPaginator with translator


petrparolek:
Version: v5.x (3c8ee74)

Example:
```
	/**
	 * @var \Kdyby\Translation\Translator
	 * @inject
	 */
	public $translator;

	public function createComponentItemsGrid($name)
	{
		$grid = new DataGrid($this, $name);

		$grid->setDataSource($datasource);
		$grid->addColumnText('name', 'Name');

		$grid->setTranslator($this->translator);
	}
```

If set translator, show all items per page doesn't work

petrparolek:
done

PR #724, opened 2018-11-07
Fix bug in DataGridPaginator with translator (#723)

petrparolek:


petrparolek:
@paveljanda @f3l1x

paveljanda:
Thank you!

PR #736, opened 2019-01-18
Add check for __toString() in array row value

rixafy:
When you tried to load array from query builder (with HYDRATE_ARRAY hydrator), and Entity has $id column of type UuidInterface or so on, it was returned in Row::getValue() instead of string (for instance in links). So I added checking for object type and __toString() function.


paveljanda:
Okey, I like that, thank you

PR #757, opened 2019-03-15
Added method ::getOptions() for FilterSelect and FilterMultiSelect.

stanislav-janu:
If I want show filters as pills like:

```
<span class="label label-primary offset-right">
	{$filter->getName()}:
	{foreach $filter->getValue() as $value}
		{$filter->getOptions[$value]}{sep}, {/sep}
	{/foreach}
</span>
```

Now is no effective way.


f3l1x:
cc @paveljanda

paveljanda:
👍

PR #759, opened 2019-03-17
Fire onCustomRedraw event also when inline edit is cancelled

machv:
Hi,
By introducing this small change `onCustomRedraw` event would be fired also when `cancel` button is pressed while in inline editing mode, right now that event is fired only when form is actually submitted, and in my use-case it would be really handy to be able to react to both situations (submit and cancel). 

It also passes as an event parameter name of the button which was used, other option would be to pass the instance of that button itself.

It will allow usage like this:
```php
$grid->getInlineEdit()->onCustomRedraw[] = function ($sourceButtonName) use ($grid) {
	if($sourceButtonName == "cancel")
		$grid->redrawControl();
};
 ``` 
Or is there any intentional behavior why is currently `onCustomRedraw` event fired only when submit `button` is used? As it might be a BC break, if it was intended only for submits by design.


paveljanda:
Could you please make a PR into the v6.x branch? Thanks a lot!

machv:
I am slightly struggling with the PR's rebasing, so I am closing this one and I created a new PR (#785) against the `v6.x` branch.

PR #760, opened 2019-03-18
Nextras Orm v2.3 under PHP 7.2

f3l1x:


f3l1x:
cc @paveljanda, cool right? :-) We are able to run with nextras/orm on php 7.2 ;-)

PR #769, opened 2019-04-08
$sort init as empty array

jirinapravnik:
in getSortNext() there is called reset() on null array


paveljanda:
@jirinapravnik Thank you!

PR #771, opened 2019-04-08
add missing Ublaboo\DataGrid\Column\Column use to TButtonRenderer

dakorpar:
PHPStan is screaming on this one on my projects :)

### DIFF
--- a/src/Traits/TButtonRenderer.php
+++ b/src/Traits/TButtonRenderer.php
@@ -8,6 +8,7 @@
 
 namespace Ublaboo\DataGrid\Traits;
 
+use Ublaboo\DataGrid\Column\Column;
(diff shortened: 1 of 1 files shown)


paveljanda:
But there is no mention of this class in that file

dakorpar:
@paveljanda it is in phpdoc comment on line 117 and this cause phpstan error afterwards and no autocomplete in IDE

 32     Call to method setFilterText() on an unknown class
         Ublaboo\DataGrid\Traits\Column.

paveljanda:
Oh I see, I am working on version 6 where there is no line number 117 🙈

f3l1x:
Is it still relevant? @paveljanda

paveljanda:
I do not think so

PR #775, opened 2019-04-13
Enable Nette v3

trejjam:
BC Break? Probably

It relaxes dependencies to enable Nette v3 dependencies. To keep Nette v3 support relatively easy it drops <PHP 7.1 support. Also, it sets minimal Nette dependencies after Object to SmartObject switch.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,8 +1,6 @@
 language: php
 
 php:
-  - 5.6
-  - 7.0
   - 7.1
   - 7.2
   - 7.3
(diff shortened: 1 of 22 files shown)


paveljanda:
Hi, thanks a lot man. Have you checked the branch `v6.x`? We are preparing a big release in this branch. Could you please compare your branch witch `v6.x`?

trejjam:
Hi, thanks. I miss the existence of `v6.x` (I checkout `master` and were doomed).

This PR does not contain anything useful what `v6.x` doesn't have already. It works well by the way :+1:

PR #776, opened 2019-04-22
Handle the invalid date values in the date and date range filters

petaak:
Handle the invalid date values in the date and date range filters. Do not apply the filter when the date string is invalid (DataGridDateTimeHelperException is thrown).


paveljanda:
Thank you, perfect

PR #777, opened 2019-04-23
Handle the invalid date values in the date and date range filters - v5.x version

petaak:
Handle the invalid date values in the date and date range filters. Do not apply the filter when the date string is invalid (DataGridDateTimeHelperException is thrown).

v5.x version of https://github.com/contributte/datagrid/pull/776


paveljanda:
Thanks!

PR #778, opened 2019-04-24
Fix typeError

MartinSzollos2016:
array_keys() expects parameter 1 to be array, object given

Doc: https://doc.nette.org/en/3.0/sessions#toc-sections
At the end of the paragraph

### DIFF
--- a/src/DataGrid.php
+++ b/src/DataGrid.php
@@ -2036,7 +2036,7 @@ s	 */
 			$this->deleteSessionData('_grid_has_sorted');
 		}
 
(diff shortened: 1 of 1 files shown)


paveljanda:
Okey, thanks

MartinSzollos2016:
ping @paveljanda 
Datagrid.php:2092

Your new code on PHP7.3  throws 
TypeError
Argument 1 passed to iterator_to_array() must implement interface Traversable, array given

Use as i commited above.
`foreach ($section as $key => $val) {
    echo "$key = $val";
}`

PR #627, opened 2018-03-12
Replacement of getName with lookupPath

peldax:
Changes:
- new protected variable in DataGrid class `componentFullName`, which contains result of `$this->lookupPath()`.
- replaced occurrences of `$this->getName()` with `$this->componentFullName` in DataGrid class and in datagrid.latte

By doing this, datagrid works with its full and unique name.

It is solution to #538 , where link was generated incorrectly, because the component's name was not full.

It might fix some other issues, because some JS selectors are currently using componenets name.

PR #787, opened 2019-05-25
Replacement of getName with lookupPath #2

peldax:
Follows https://github.com/contributte/datagrid/pull/627

New branch & rebase.


peldax:
Ping @paveljanda.

This is the new PR with clean branch and rebase.

paveljanda:
Thank you. There is just one problem: When you define the datagrid like so:

```php
$grid = new DataGrid;

#...

return $grid
```
then the parent is not yet defined. In another words, the `DataGrid` component is not yet attached to the presenter so the `$this->lookupPath()` will throw an exception: `Component of type 'Ublaboo\DataGrid\DataGrid' is not attached to ''.`.

peldax:
Hmm, if we move the `lookupPath()` call from constructor to monitor callback, is there any possible issue?

Or we could lazy load the `componentFullName` property in getter function `getFullName()`.

paveljanda:
This is a good idea.
I would definitely move it to the monitor callback + call `::getFullName()` from everywhere instead of using the `::$componentFullName` property directly. :) + I could you maybe add a `Exception` throw to the `::getFullName()` method when the component has `::$componentFullName === null` (the component is not attached to the presenter yet)? Thanks man. :) I appreciate it

peldax:
I noticed that lookupPath was originally inside monitor (otherwise it couldn't work, as you said), but got moved one level up because of bad merge, sorry about that.

I also noticed you are calling monitor function twice in constructor, is it intented?

paveljanda:
Thank you! :)

paveljanda:
https://github.com/contributte/datagrid/releases/tag/v6.0.1

PR #789, opened 2019-05-27
Render condition on status column

peldax:
This PR adds ability to restrict rendering of status column (eg. on rows without proper permission).

Renders only text, without interactive button, if condition is not met.


paveljanda:
I like it, thanks. :)

PR #796, opened 2019-06-04
Datagrid paginator - 'all' option fix

### LINKED ISSUE #795: paginator option 'All' does not work - v5.7.2
Hello, i have problem with paginator when is selected All option. Page is refreshed, but displayed values are from first option of Paginator (default 10).

Using latest Nette 2.4 and Dibi 4. Data source is Dibi Fluent. 

I found problem in method `getPerPage` at class ` \Ublaboo\DataGrid\DataGrid`:
```
public function getPerPage()
	{
		$items_per_page_list = $this->getItemsPerPageList();
		$per_page = $this->per_page ?: reset($items_per_page_list);

		if (($per_page !== 'all' && !in_array((int) $this->per_page, $items_per_page_list, true)) // there should be array_keys($items_per_page_list)
			|| ($per_page === 'all' && !in_array($this->per_page, $items_per_page_list, true))) { // there should be array_keys($items_per_page_list)
			$per_page = reset($items_per_page_list);
		} 
		return $per_page;
	}
```

Thank you!
EDIT: 
Problem is when values are translated. The condition looking into values (instead of keys) and searching 'all', but there isnt one. 
image

BigOHenry:
#795

### DIFF
--- a/src/DataGrid.php
+++ b/src/DataGrid.php
@@ -2783,8 +2783,8 @@ class DataGrid extends Nette\Application\UI\Control
 
 		$per_page = $this->per_page ?: reset($items_per_page_list);
 
-		if (($per_page !== 'all' && !in_array((int) $this->per_page, $items_per_page_list, true))
-			|| ($per_page === 'all' && !in_array($this->per_page, $items_per_page_list, true))) {
+		if (($per_page !== 'all' && !in_array((int) $this->per_page, array_keys($items_per_page_list), true))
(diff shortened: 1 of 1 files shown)


BigOHenry:
Ah, i am blind... There is already fix -` $items_per_page_list = array_keys($this->getItemsPerPageList());`

But it isnt at newest 5.7.2 version.

paveljanda:
Thanks, I will have a look at the releases and release 5.7.3 when there is a fix in v5.x. 👍

BigOHenry:
Thank you :)

PR #800, opened 2019-06-17
Support inheritance static analysis in return types

mabar:


paveljanda:
Awesome, thanks a lot!

Issue #803, opened 2019-06-21
StatusColumn doesn't work


BigOHenry:
Hello,
i have this array for column status:
![image](https://user-images.githubusercontent.com/15823552/59906537-3cf7cc80-9409-11e9-8b5b-95f9acc50d79.png)

Result in datagrid:
![image](https://user-images.githubusercontent.com/15823552/59906620-7a5c5a00-9409-11e9-8b39-c27651af9d28.png)

Problem is method `getCurrentOption` in `Ublaboo\DataGrid\Column\ColumnStatus` class:
```
        /**
	 * Find selected option for current item/row
	 */
	public function getCurrentOption(Row $row): ?Option
	{
		foreach ($this->getOptions() as $option) {
           // bdump($option->getValue());
           // bdump($row->getValue($this->getColumn()));
           // bdump('--------');
			if ($option->getValue() === $row->getValue($this->getColumn())) {
				return $option;
			}
		}

		return null;
	}

```
There is a dump:
![image](https://user-images.githubusercontent.com/15823552/59906732-b4c5f700-9409-11e9-8525-29d110545044.png)

Possible fix:
**1)** 
```
if ($option->getValue() == $row->getValue($this->getColumn())) {
	return $option;
}
```
OR
**2)**
```
if (strval($option->getValue()) === strval($row->getValue($this->getColumn()))) {
	return $option;
}
```
OR 
**3)** 
`$option->getValue()` should return string

BigOHenry:
I made PR for option **2)**

BigOHenry:
What is status about this issue? Maybe better solution would be option 2 or 3.

BigOHenry:
I made fix for this in #827, old communication is at #804.

BigOHenry:
Any motivation fixing this? I made PR #827 almost year ago with no reaction. Thank you

paveljanda:
I will look into it later this day

paveljanda:
It is likely caused by your database driver, with dibi, it works fine.

What database driver are you using?

When your driver displays `"1"` for `1`, please set the options using:

```php
->setOptions([
	'4' => 'Plánovaná',
	'5' => 'Dokončená',
	'6' => 'Zrušená',
]);
```

BigOHenry:
Thank you.
I am using dibi too. This is not exact problem. I will send more info
tomorow.

Dne st 24. 6. 2020 16:12 uživatel Pavel Janda <notifications@github.com>
napsal:

[...]

PR #804, opened 2019-06-21
Fixed column status

BigOHenry:
Fix for https://github.com/contributte/datagrid/issues/803

### DIFF
--- a/src/Column/ColumnStatus.php
+++ b/src/Column/ColumnStatus.php
@@ -91,7 +91,7 @@ class ColumnStatus extends Column
 	public function getCurrentOption(Row $row): ?Option
 	{
 		foreach ($this->getOptions() as $option) {
-			if ($option->getValue() === $row->getValue($this->getColumn())) {
+			if ($option->getValue() == $row->getValue($this->getColumn())) {
 				return $option;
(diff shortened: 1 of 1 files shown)


paveljanda:
Actually, it works fine for me. What data source do you use?

BigOHenry:
Just tried version 6.2.0 and problem still persist. Data source is `Dibi\Fluent`.

BigOHenry:
Problem still persist in 6.2.x.

src/Column/ColumnStatus.php
```diff
@@ -91,7 +91,7 @@ public function getColumn(): string
 	public function getCurrentOption(Row $row): ?Option
 	{
 		foreach ($this->getOptions() as $option) {
-			if ($option->getValue() === $row->getValue($this->getColumn())) {
+			if ($option->getValue() == $row->getValue($this->getColumn())) {
```
mabar: Check for exact "empty" values would be better than ==

BigOHenry: I made new PR because i deleted that old fork. https://github.com/contributte/datagrid/pull/827

PR #808, opened 2019-07-16
Refactoring (repo, composer, doctrine)

f3l1x:


f3l1x:
Ready for review. @paveljanda

.scrutinizer.yml
```diff
@@ -6,5 +6,5 @@ tools:
 
 filter:
   excluded_paths:
-    - docs/*
+    - .docs/*
```
paveljanda: 👍

src/DataSource/DoctrineCollectionDataSource.php
```diff
@@ -4,9 +4,10 @@
 
 namespace Ublaboo\DataGrid\DataSource;
 
-use Doctrine\Common\Collections\ArrayCollection;
 use Doctrine\Common\Collections\Collection;
 use Doctrine\Common\Collections\Criteria;
+use Doctrine\Common\Collections\Selectable;
+use LogicException;
```
paveljanda: Could you please refer to native classes directly in code and not in the use statement? Like `\LogicException`. Reason is, there are many vendors using the same class name as the native one (for example, `Dibi\DateTime`) and thus when you use '\DateTime`, you can see what class it is right away.

Than you. 👍

f3l1x: Yup

PR #809, opened 2019-07-23
Handle DataGrid::getSessionData returning plain array

trejjam:
DataGrid with `$rememberState = false` may returns plain array from `getSessionData` and not only `Traversable`. This produce an error calling `handleResetFilter`.

This PR solves the issue and introduce missing tests.


trejjam:
Hi, the reported bug was fixed by someone else. If you do not want to append these tests, feel free to close the PR.

f3l1x:
Tests are fine ;-)

PR #810, opened 2019-07-24
CreateComponent bubbles to parent component

peldax:
Feature
BC break - probably not

I added createComponent($name) method, which bubbles to parent if component is not found in current class.

This change is useful eg to include component into itemDetail and/or other templates. Templates belong to Datagrid, so there is current no way to use {control} macro in templates without overriding Datagrid.

### DIFF
--- a/src/DataGrid.php
+++ b/src/DataGrid.php
@@ -3313,4 +3313,16 @@ class DataGrid extends Control
 		return $presenter;
 	}
 
+	
+	protected function createComponent(string $name): ?\Nette\ComponentModel\IComponent
+	{
+		$control = parent::createComponent($name);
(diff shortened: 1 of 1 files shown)


f3l1x:
Hi @peldax. Could you please add some tests? I don't know what side effects it could have.

peldax:
Hi,
I could have back then, but its been more than 2 years and we no longer use this datagrid.
Feel free to close this PR.

f3l1x:
What are you using instead of it? Just curious.

peldax:
In projects with generated frontend we opt to use our private fork. Otherwise we have frontend/client in completely different technology.

PR #811, opened 2019-07-31
Fix paginator for translated "All" (#795)

romanmatyus:
Check key for paginator, not translated value.


paveljanda:
Done already, thanks

PR #813, opened 2019-08-06
Added the casting to string to the CsvDataModel::getRow()

jaroslavlibal:
- strip_tags() function expects parameter 1 to be string


paveljanda:
Thank you 👍

PR #815, opened 2019-08-08
Make the $template parameter nullable in Column::setTemplate()

jaroslavlibal:
Lets you turn off the template for columns with a default template.


paveljanda:
Thanks!

PR #818, opened 2019-08-12
Cast string on values

MartinSzollos2016:
`$grid->addColumnNumber('test', 'Test')
			->setFilterSelect(
				[
					0 => '0',
					1 => '1',
				]
			);`

When u try to filter u get:
TypeError
explode() expects parameter 2 to be string, int given


f3l1x:
@paveljanda

paveljanda:
@MartinSzollos2016 Could you maybe make a type cast for the explode function?

paveljanda:
Okey, cool, let's do that. Thanks

PR #819, opened 2019-08-12
Pass UI and event into nette ajax call

zipper:
`settings.nette` object built in `nette.ajax.js` requires UI and event to be passed. Otherwise, there is no `settings.nette` available in nette extensions. This commit changes `$.nette.ajax()` calls in sortable callbacks so the handle is passed as UI and sortable event as event. Thanks to this, it is possible to eg. make spinner work when sorting is being processed.


paveljanda:
Cool, thanks.

PR #821, opened 2019-08-14
Add posibility to override/extend html class for the containing div

Jakub-Fajkus:
It would be nice to be able to add a custom html class to the containing div of the whole grid. In our use case, we would use that to add a html class to properly display a spinner when an user clicks the `Filter` button or uses the drag&drop sorting of the rows.

### DIFF
--- a/src/templates/datagrid.latte
+++ b/src/templates/datagrid.latte
@@ -14,7 +14,7 @@
  * @param InlineEdit|null $inlineAdd   Inline add data
  *}
 
-<div class="datagrid datagrid-{$control->getFullName()}" data-refresh-state="{link refreshState!}">
+<div class="{block datagrid-class}datagrid datagrid-{$control->getFullName()}{/block}" data-refresh-state="{link refreshState!}">
 	<div n:snippet="grid">
 	{snippetArea gridSnippets}
 		{form filter, class => 'ajax'}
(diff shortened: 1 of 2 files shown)


f3l1x:
👍wdyt @paveljanda?

src/templates/datagrid.latte
```diff
@@ -14,7 +14,7 @@
  * @param InlineEdit|null $inlineAdd   Inline add data
  *}
 
-<div class="datagrid datagrid-{$control->getFullName()}" data-refresh-state="{link refreshState!}">
+<div class="{block grid-class}datagrid datagrid-{$control->getFullName()}{/block}" data-refresh-state="{link refreshState!}">
```
paveljanda: Maybe pls `{block datagrid-class}`? (There probably more references to a "datagrid" than to a "grid"). Thanks

Jakub-Fajkus: Sure, everything is called datagrid... changed that

paveljanda:
Nice thanks

paveljanda:
@Jakub-Fajkus Can you please update also the docs? Add somewhere that this feature is available. Thx

Jakub-Fajkus:
@paveljanda Added a new section `Containing div class definition` (not sure about the name, though). I also added the missing class to section `Table class definition`.

paveljanda:
Thanks!

PR #824, opened 2019-08-23
Add method for setting custom Action href

Gappa:
Do you still accept PRs for 5.x branch? 

I would also like this feature for 6.x, but still haven't updated to Nette 3, so that will have to wait :)

### DIFF
--- a/src/Column/Action.php
+++ b/src/Column/Action.php
@@ -87,6 +87,11 @@ class Action extends Column
 	 */
 	private $title;
 
+	/**
+	 * @var string|callable
+	 */
+	private $customHref;
+
 
 	/**
 	 * @param DataGrid $grid
@@ -121,11 +126,15 @@ class Action extends Column
 			// Do not use renderer
 		}
 
-		$link = $this->createLink(
-			$this->grid,
-			$this->href,
-			$this->getItemParams($row, $this->params) + $this->parameters
-		);
+		if (!empty($this->customHref)) {
+			$link = $this->getCustomHref($row);
+		} else {
+			$link = $this->createLink(
+				$this->grid,
+				$this->href,
+				$this->getItemParams($row, $this->params) + $this->parameters
+			);
+		}
 
 		$a = Html::el('a')->href($link);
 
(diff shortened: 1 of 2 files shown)


dakorpar:
Wouldn't better implementation be to check if passed argument is absolute url and use it with already existing parameter href?

Gappa:
Well, it might not be an absolute URL (with protocol and everything), it might be relative - pointing to some other part of the application.

Also, href argument can be empty or just `action` - the `createLink` method then does its magic, so I didn't want to interfere with that in any way.

Maybe I didn't understand your point correctly? :)

dakorpar:
hm, if it's relative, then it *should* be covered with how it works now, not sure about empty href, but I see what you mean...

Gappa:
It is covered, but in a basic way - the parameters are static and if it is a computed value, you need to have a method for that in your entity (if the datasource is that) which isn't all that intuitive at first.

And for conditional cases it is currently afaik impossible - something like this:

``` php
->setCustomHref(function(Entity $entity) use ($presenter) {
	if ($entity->getSometing()) {
		return $presenter->link(':Admin:Presenter:action', ['param' => $entity->getSometing()])
	} else {
		return 'http://www.example.com/?id=' . $entity->getId();
	}
});
```

This should be the typical use-case for this method.

I should have explained the PR more thoroughly in the initial comment.

Gappa:
Hi, I would like to ask about this PR:
- Is it even feasible to add features to 5.x branch?
- Is the code itself acceptable, or needs polishing?
- Is the feature something that shouldn't be included in the package? If that is the case, that's ok :)

As a workaround, I currently use complete manual rendering, but that takes more code when you want the button/icon combo to remain. Maybe there's a better way and this whole PR is meaningless? :)

Anyway, tests fail with this message (but should otherwise pass):

- `The command "mysql -e 'create database datagridTest;'" failed and exited with 1 during.` 

Maybe they stoppding working for the 5.x branch?

paveljanda:
Thanks, I like it.

- Yes, it is. But could you please prepare also a MR into the master branch? I'd like to keep this feature also in v6. Thanks!
- It's fine. :)
- It sure is. :)

Gappa:
Ok, cool :)

Made PR for the v6.x branch - #853

f3l1x:
Can we close it here?

Gappa:
> Can we close it here?

If there is no plan to include it in 5.x, then I think we can :)

PR #849, opened 2019-11-25
fix default value in small inline edit

KrejcirD:
This changes the small inline edit to set the default value via the value and not the text of the select box.


paveljanda:
Thank you!

PR #850, opened 2019-11-25
V5.x - merge v5.7.3 and fix array_key_exists in getPerPage method

xsuchy09:
Merge v5.7.3 into 5.x branch and fix array_key_exists - this function has only 2 parameters, not 3.

### DIFF
--- a/src/DataGrid.php
+++ b/src/DataGrid.php
@@ -2783,8 +2783,7 @@ class DataGrid extends Nette\Application\UI\Control
 
 		$per_page = $this->per_page ?: reset($items_per_page_list);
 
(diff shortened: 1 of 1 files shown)


xsuchy09:
I just change tag v5.7.3 position to the new commit with this fix (with -f param) ...

paveljanda:
Actually, your MR removes a bit much of the original code. I fixed in in the v5.x branch. But thanks!

xsuchy09:
> Actually, your MR removes a bit much of the original code. I fixed in in the v5.x branch. But thanks!

Yes I merged v5.7.3 before my changes ... it shouldn't be there?

xsuchy09:
There is still an issue - see https://github.com/contributte/datagrid/pull/858

PR #855, opened 2019-11-28
The reset column filter functionality is broken when the datagrid is nested inside another component

Jakub-Fajkus:
The reset column filter functionality is broken, when the datagrid component is nested inside another component.

Just to be clear, it is the X link in the table heading (screen included)

reset_column

So, the proposed solution is as follows.
I've removed the JS code, that changes the URL of the link to delete the filter on a column. The comment said that the code is there because table header is not refreshed using snippets. So, I've added the snippet to the `thead` element. Next, the snippet is invalidated in the `\Ublaboo\DataGrid\DataGrid::reload()` method.

Basically, I wonder why it was originally done that way and not via the snippet invalidation.

I've tested that change for grid, which is directly attached to the presenter as well as grid inside another component and it seems to be working just fine. i don't think this could cause any issue other than slightly bigger network traffic.

### DIFF
--- a/assets/src/datagrid.coffee
+++ b/assets/src/datagrid.coffee
@@ -722,16 +722,4 @@ $.nette.ext('datagrid.reset-filter-by-column', {
 		if payload.non_empty_filters && payload.non_empty_filters.length
 			for key in payload.non_empty_filters
 				grid.find('[data-datagrid-reset-filter-by-column='+key+']').removeClass('hidden')
-
-			# Refresh their url (table header is not refreshed using snippets)
-			#
-			href = grid.find('.reset-filter').attr('href')
-
-			grid.find('[data-datagrid-reset-filter-by-column]').each ->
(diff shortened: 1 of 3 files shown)


Jakub-Fajkus:
@paveljanda @f3l1x What do you think? I would love to see that in the 5.7.* versions since I consider that a bug and because there are still people (inluding me) who cannot yet update to nette 3.0.

f3l1x:

paveljanda:
All this code was added because older (I am not sure whether it still happens in newer versions) nette versions did not behave exactly as intended - speaking about snippets. To merge this PR, it would require a thorough testing across multiple nette versions and all possible scenarious with both this and without this PR. :/

f3l1x:
Please rebase your PR if it's still relevant.

f3l1x:
Closing here, feel free to reopen it.

PR #860, opened 2019-12-16
Fix sort callback condition

vitkutny:
After fec69ac3aa298fa86c03b0b86af93a466340ab11 `!$sort_callback` has been incorrectly changed to `$sortCallback !== null`

### DIFF
--- a/src/DataGrid.php
+++ b/src/DataGrid.php
@@ -833,7 +833,7 @@ class DataGrid extends Control
 			$sort[$column->getSortingColumn()] = $order;
 		}
 
-		if ($sortCallback !== null && isset($column)) {
+		if ($sortCallback === null && isset($column)) {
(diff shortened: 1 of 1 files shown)


mskocik:
@f3l1x @paveljanda This should be merged immediately. Because without it paging doesn't work for column with `sortableCallback` defined and given column is being used for sorting.

paveljanda:
Could you please complete some tests to prevent broking this functionality? Thanks a lot!

mskocik:
> Could you please complete some tests to prevent broking this functionality? Thanks a lot!

@vitkutny  Can you do it?

vitkutny:
I dont have time for this, sorry. Its just bugfix of simple mistake.
Screenshot 2020-05-13 at 12 14 52

radimvaculik:
@paveljanda Please, close this PR. Already fixed.

PR #861, opened 2019-12-17
Fix context in naja and $.nette adapters

vitkutny:
Because these functions are defined in global scope their "this" value is window. When we assign another function to this variables, their context is not changed. If any function uses ($.nette.ajax does) "this" it will not work because "this" will be "Window" instead of expected object.

### DIFF
--- a/assets/datagrid.js
+++ b/assets/datagrid.js
@@ -51,11 +51,19 @@ if (typeof naja !== "undefined") {
 			.catch(params.error);
 	};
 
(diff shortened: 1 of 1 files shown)


ikeblaster:
@vitkutny Same problem is also in `datagrid-instant-url-refresh.js`, can you please append fix to this PR?

paveljanda:
Thanks!

paveljanda:
@ikeblaster Great point 👍

PR #862, opened 2019-12-18
Fixed creating item detail form container after submit

vitkutny:
Fixes `Component with name 'content' does not exist.` when rendering detail form after submit. Its implemented similiar to `\Nette\Application\UI\Multiplier::createComponent`. Its simplified and will also work when container is accessed using `final public function getComponent` and not only using array access `offsetGet`.


f3l1x:

paveljanda:
I am not sure about it. This code was added so that it fixes some behaviour or old nette versions. As there are no tests (my bad :/), I can not be sure that it won't break old nettes.

f3l1x:
Right, @vitkutny can you provide some tests?

PR #863, opened 2019-12-18
Fix items detail form parameter in detail template

vitkutny:
Item detail form parametr is not available after update to current version of datagrid.

In `Datagrid.php` item detail form is set as `items_detail_form ` and not `itemsDetail_form `.
```
		if ($itemsDetailForm instanceof Container) {
			$form['items_detail_form'] = $itemsDetailForm;
		}
```


paveljanda:
Uuh, thanks!

Issue #864, opened 2019-12-19
Can not set default filter for FilterDateRange


xsuchy09:
When you are trying to set default filter for FilterDateRange it is ending with:

> Ublaboo\DataGrid\Exception\DataGridException
> Filter [from] is not defined

It is because of calling `$this->getFilter` in `Datagrid::setFilterContainerDefaults`:
```
			if ($value instanceof \DateTime) {
				$filter = $this->getFilter($key);

				if ($filter instanceof IFilterDate) {
					$value = $value->format($filter->getPhpFormat());
				}
			}
```

`$key` is here `from` (and `to`) so it can not be found in filters ... it has to be changed (called with parent key name) - see pull request.

**Steps to reproduce:**
1. Create DataGrid with DateTime column and FilterDateRange filter - for example `date`:
```
$dataGrid->addColumnDateTime('date', 'Date')
	->setFilterDateRange();
```
2. Set default values for that column - for example:
```
$dataGrid->setDefaultFilter([
	'date' => [
		'from' => (new \DateTime())->sub(new \DateInterval('P3M')),
		'to' => (new \DateTime())->sub(new \DateInterval('P1D')),
	]
]);
```

It will crash with:
> Ublaboo\DataGrid\Exception\DataGridException
> Filter [from] is not defined

xsuchy09:
Up ... still in a new version too ...

paveljanda:
Will be in v6.2.15

PR #865, opened 2019-12-19
Fix: Can not set default filter for FilterDateRange #864

xsuchy09:
*fix - default filter for FilterDateRage can not be set
*issue - https://github.com/contributte/datagrid/issues/864

### DIFF
--- a/src/DataGrid.php
+++ b/src/DataGrid.php
@@ -1501,7 +1501,7 @@ class DataGrid extends Control
 	}
 
 
-	public function setFilterContainerDefaults(Container $container, array $values): void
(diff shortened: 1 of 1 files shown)


paveljanda:
I don't really understand, how is this broken. See https://examples.contributte.org/packages/datagrid/filters?grid-perPage=20&grid-filter%5Bbirth_date_2%5D%5Bto%5D=7.%204.%202020

paveljanda:
Aha! I see, will fix that

paveljanda:
Awesome, thanks!

PR #868, opened 2020-01-13
datagrid-instant-url-refresh.js: dropped jQuery dependency

martenb:


paveljanda:
That's great, thanks!

martenb:
@paveljanda 
Perhaps it would be good to split datagrid.js into smaller parts and then gradually remove the dependency on jQuery. What do you think?

PR #869, opened 2020-01-17
Multiple AggregationFunction fix

dakorpar:
Potential BC for users who wrote their own aggregation functions but also bugfix since IMultipleAggregationFunction.php could not work at all now
This was broken on few layers, this PR is fixing it

### DIFF
--- a/src/AggregationFunction/FunctionSum.php
+++ b/src/AggregationFunction/FunctionSum.php
@@ -88,7 +88,7 @@ class FunctionSum implements IAggregationFunction
 	/**
 	 * @return mixed
 	 */
-	public function renderResult()
+	public function renderResult(?string $key = null)
 	{
 		$result = $this->result;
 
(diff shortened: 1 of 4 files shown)


paveljanda:
Thanks for the PR. I fixed it in #870.
The `$key` parameter is not needed though. 

https://github.com/contributte/datagrid/releases/tag/v6.2.8
https://github.com/contributte/datagrid/blob/master/.docs/columns.md#aggregation-function

dakorpar:
@paveljanda  I think you got it wrong....
I really want to use IMultipleAggregationFunction which was also supported in previous version and rendered with key:

https://github.com/dakorpar/datagrid/blob/84c1e5ff88f554c0689736462bb033ac40580a43/src/templates/datagrid.latte#L491

IAggregationFunction will never work with multiple columns since it queries database for every column and when having big and complicated query that just won't work because it will be to slow.

paveljanda:
You are right. Resolved in PM and version `6.2.9`:

https://github.com/contributte/datagrid/releases/tag/v6.2.9
https://github.com/contributte/datagrid/blob/master/.docs/columns.md#aggregation-function

PR #875, opened 2020-02-20
GroupTextareaAction - \Nette\Forms\Container::addTextArea() require first parameter as string

VaclavPavek:


paveljanda:
Thanks!

PR #876, opened 2020-02-21
Fix inline edit request method in Naja

Darneus:
changed ajax call argument for inline edit from "method" to "type"

### DIFF
--- a/assets/datagrid.js
+++ b/assets/datagrid.js
@@ -652,7 +652,7 @@ $(document).on('click', '[data-datagrid-editable-url]', function(event) {
 					data: {
 						value: value
 					},
-					method: 'POST',
+					type: 'POST',
 					success: function(payload) {
 						if (cell.data('datagrid-editable-type') === 'select') {
 							cell.html(input.find("option[value='" + value + "']").html());


paveljanda:
@dakorpar Could you please verify that this update works? I know only that you are using Naja. Thanks a lot!

dakorpar:
@paveljanda works in both cases for me, but from code this PR looks correct, not sure why is working correct for me even now, it sends post request with this code and current release.
IMHO this should be merged

paveljanda:
@dakorpar Thanks for the info!
@Darneus Could you please provide some context why this does not work? Some documentation link or something.. Thanks

Darneus:
@paveljanda it does not work because naja implementation dataGridRegisterAjaxCall maps param "type" to "method" or use "GET" by default, but inline editing provided a "method". It's on line 46 in datagrid.js

dakorpar:
@paveljanda this is same as commit you've merged already:
https://github.com/contributte/datagrid/commit/8c99c500c10bae05ea0d6a834b4fac72fae111bd
so you can close this

paveljanda:
Cool, thanks both!

Issue #877, opened 2020-02-21
ArrayDataSource SelectBox filter fails if values are integers


kuchy:
ArrayDataSource SelectBox filter not work if values are integers, because explode gets only string and strict_types are enabled.

paveljanda:
Could you send a MR?

kuchy:
I will submit....hope that its will be today :)

PR #878, opened 2020-02-21
ArrayDataSource SelectBox filter fails if values are integers

kuchy:
Fix for https://github.com/contributte/datagrid/issues/877


paveljanda:
Thanks!

PR #881, opened 2020-03-10
DibiFluentPostgreDataSource - fix for BC break in dibi 4.1.2 in text search

elring:
In Datagrid with DibiFluentPostgreDataSource text filters don't work with recent version of Dibi.

Reason is Dibi 4.1.2, where [dibi PR#346](https://github.com/dg/dibi/commit/294787a26e8bafdb6622d17300dbb8194b552304) changes meaning of function escapeLike parameter pos (0 => bitmask).

This PR simply changes escaping to not-use this function like others DibiFluent*DataSources do. So the code works with any version of Dibi (pre or post 4.1.2).


adaamz:
I think we still need to wait for new stable version of tharos/leanmapper, because current stable version (3) supports only dibi 3.

paveljanda:
@adaamz How is that related to leanmapper?

adaamz:
Yeah, probably not much related.

paveljanda:
Ok, thanks

PR #883, opened 2020-03-14
Upgraded libraries, synced CI with contributte/bare

adaamz:
- Upgraded PHPStan to version 0.12
- Removed useless gamee phpstan rules (which is only phpstan-strict-rules)
- Added whole ninjify/qa
- Synced travis-ci with contributte/bare


adaamz:
I allowed failure of php 7.4 tests, because tharos/leanmapper has no stable version for php 7.4.

Questionable is why we need to install tharos/leanmapper - it looks like there is no usage (neither in tests) of this lib.

f3l1x:
LGTM! What about @paveljanda?

src/DataSource/DibiFluentDataSource.php
```diff
@@ -60,7 +60,9 @@ public function getCount(): int
 	 */
 	public function getData(): array
 	{
-		return $this->data ?: $this->dataSource->fetchAll();
+		return $this->data !== []
+			? $this->data
+			: $this->dataSource->fetchAll();
```
mabar, line 65: Shouldn't that also write into $this->data, if it's empty?
Maybe property should also be null by default to distinguish between uninitialized and no data states

src/DataSource/NetteDatabaseTableDataSource.php
```diff
@@ -82,7 +82,9 @@ public function getCount(): int
 	 */
 	public function getData(): array
 	{
-		return $this->data ?: $this->dataSource->fetchAll();
+		return $this->data !== []
+			? $this->data
+			: $this->dataSource->fetchAll();
```
mabar, line 87: Same as DibiFluentDataSource::getData()

src/DataSource/DibiFluentDataSource.php
mabar, line 65: Shouldn't that also write into $this->data, if it's empty?
Maybe property should also be null by default to distinguish between uninitialized and no data states

src/DataSource/NextrasDataSource.php
```diff
@@ -63,7 +63,9 @@ public function getData(): array
 		/**
 		 * Paginator is better if the query uses ManyToMany associations
 		 */
-		return $this->data ?: $this->dataSource->fetchAll();
+		return $this->data !== []
+			? $this->data
+			: $this->dataSource->fetchAll();
```
mabar, line 68: Same as DibiFluentDataSource::getData()

src/Export/ExportCsv.php
```diff
@@ -46,21 +45,20 @@ private function getExportCallback(
 			array $data,
 			DataGrid $grid
 		) use ($name, $outputEncoding, $delimiter, $includeBom): void {
-			$columns = $this->getColumns() ?: $this->grid->getColumns();
+			$columns = $this->getColumns();
+			if ($columns === []) {
+				$columns = $this->grid->getColumns();
+			}
```
mabar, line 51: Looks like getColumns() and grid->getColumns() returns the same, shouldn't getColumns() call and cache grid->getColumns() internally? cc @paveljanda

f3l1x:
@adaamz Here is some test for leanmapper (https://github.com/contributte/datagrid/blob/42bcd36fc2d03bbee3fc4cad7e04864557c8f05d/tests/Cases/RowTest.phpt#L84-L94).

f3l1x:
WDYT @paveljanda? I think we can merge it.

src/DataModel.php
```diff
@@ -76,9 +76,6 @@ public function __construct($source, string $primaryKey)
 			if ($driver instanceof Dibi\Drivers\OdbcDriver) {
 				$source = new DibiFluentMssqlDataSource($source, $primaryKey);
 
-			} elseif ($driver instanceof Dibi\Drivers\MsSqlDriver) {
-				$source = new DibiFluentMssqlDataSource($source, $primaryKey);
-
```
paveljanda, line 81: Why is this code removed?

src/Export/ExportCsv.php
paveljanda: I don't think so. There could be custom columns in ExportCsv set.

src/Filter/Filter.php
```diff
@@ -236,7 +236,7 @@ protected function addAttributes(BaseControl $input): BaseControl
 				$value = implode(' ', $value);
 			}
 
-			$input->setAttribute($key, $value);
+			$input->setHtmlAttribute($key, $value);
```
paveljanda, line 239: 👍

src/DataModel.php
adaamz: https://github.com/dg/dibi/releases/tag/v4.0.0 mssqldriver was removed in dibi 4.0.0

paveljanda: Then is should be maybe replaced with SQLSRV driver? (And tested if replacing)

paveljanda: We can't just drop this support, it would be a BC Break.

composer.json
```diff
@@ -36,14 +36,16 @@
...
 		"nette/database": "~3.0.2",
 		"nette/tester": "^2.2",
 		"nextras/orm": "~3.1.0",
-		"ninjify/coding-standard": "^0.9.0",
-		"phpstan/phpstan": "^0.11.12",
-		"phpstan/phpstan-strict-rules": "^0.11.1",
+		"ninjify/qa": "^0.10.0",
+		"phpstan/phpstan": "^0.12.2",
+		"phpstan/extension-installer": "^1.0",
+        "phpstan/phpstan-deprecation-rules": "^0.12.0",
```
paveljanda, line 46: pls tabs, thx

src/DataModel.php
adaamz: Condition for SqlsrvDriver is few lines below, so I think its fine.

I added class_exists check, hopefully will be sufficient for phpstan :-)

src/DataSource/DibiFluentDataSource.php
adaamz: I'm not sure what you want from me. Writing is done in `limit` method.

adaamz:
@paveljanda @f3l1x @mabar
I
- rebased onto master
- hopefully fixed issues from CR

paveljanda:
@adaamz Could you pls resolve the conflicts? Thanks!

adaamz:
@paveljanda rebased onto master

f3l1x:
Thank you @adaamz, we have applied most of your work. Feel free to open another PR. Smaller ones are better ones.

PR #884, opened 2020-03-17
Naja support imporvement

n3t:
This patch is solving 2 issues

1/ Naja ajax search using text field filter
```
$grid->addFilterText('name', 'Name')
```
Using this filter before patch will submit whole page, after patch works as expected.

2/ Naja action confirmation
```
$grid->addAction('delete', '', 'delete!')
	->setConfirmation(
		new StringConfirmation('Really?')
	);
```
Before patch no confirmation is displayed at all, action is directly executed, after patch works as expected.


paveljanda:
Thank you!

camohub:
Can you make a tag please?

paveljanda:
https://github.com/contributte/datagrid/releases/tag/v6.2.11

camohub:
It is not updated in NPM https://www.npmjs.com/package/ublaboo-datagrid. There is still v6.2.10 I dont use Bower. Anybody shoud use it any more. Please make patch for npm not only for Bower https://docs.npmjs.com/misc/developers#publish-your-package

camohub:
???

camohub:
Are you there? Why did you make a tag only for Bower?

darkWolf-PR:
Hi, I have updated datagrid on actual version(from 6.2.11 to 6.2.17) and I still have problem with not working StringConfirmation on delete action - alert shows off, but no matter what I click, item always gets deleted. Cancel doesn´t  work/get recognized. Using Naja 1.7. Other js stuff in Datagrid like filters etc. works as expected.

n3t:
@darkWolf-PR Did you upgrade also your datagrid.js? Note that upgrading datagrid using composer will not automatically upgrade your js files.

darkWolf-PR:
Assets dir gets updated with composer no? Took new files from it. Size was different, did not actually compared em tho. Version numbers in assets files would be nice. It seems there is some problem with that confirms #900

PR #890, opened 2020-03-24
Update DataGrid.php When sortCallback is not set. Use column sortableCallback

VojtechSejkora:
- this will help when you use callback for sorting and also want to filter something.


mskocik:
Duplicate of #890. Maybe if you add test to your PR, it will get merged.

paveljanda:
Aha, thanks!

PR #891, opened 2020-04-15
Fix group action button stays invisible (2nd try)

elring:
CZ: Oprava chyby, kdy se nikdy nezobrazí tlačítko "Provést" u hromadných akcí (po kliknutí na nějakou položku ve výběru akce).

Kombinace datagrid.latte a (Nette 3.0) netteForms.js. Na počátku má tlačítko Provést v šabloně nastaveno display:none s tím, že se později zobrazí. Jenže netteForms.js (funkce Nette.toggle) to dříve dělalo pomocí smazání/nastavení atributu display, ale nově to dělá pomocí property hidden (commit), takže se nikdy nezobrazí.

### DIFF
--- a/src/templates/datagrid.latte
+++ b/src/templates/datagrid.latte
@@ -75,11 +75,11 @@
 									{='ublaboo_datagrid.group_actions'|translate}:
 									{foreach $filter['group_action']->getControls() as $form_control}
 										{if $form_control instanceof \Nette\Forms\Controls\SubmitButton}
-											{input $form_control, class => 'btn btn-primary btn-sm', style => 'display:none'}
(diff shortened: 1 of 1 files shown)


f3l1x:
Great, thx.

dakorpar:
@f3l1x @elring I would be careful with this...
Although this is BC it does not really matter since this never worked on nette3
 if someone has css where he has css style for input like display: something, this won't really work for him https://github.com/twbs/bootstrap/issues/15274

for example for me now everything is visible without even selecting anything:

image

adding
`[hidden] { display: none !important; }`
do not work either since then everything is hidden but it doesn't show on change

solution that worked for us was to add javascript listener for on change
and manually setting display style of "group_action[submit]" button

PR #897, opened 2020-04-28
Fix inline edit default value in <select>.

stanislav-janu:


paveljanda:
Current behaviour does not work?

stanislav-janu:
Current not working and my change need less code for same thing.

assets/datagrid.js
```diff
@@ -643,11 +643,7 @@ $(document).on('click', '[data-datagrid-editable-url]', function(event) {
 			input.attr('rows', Math.round(cell_lines));
 		} else if (cell.data('datagrid-editable-type') === 'select') {
 			input = $(cell.data('datagrid-editable-element'));
-			input.find('option').each(function() {
-				if ($(this).val() === valueToEdit) {
```
stanislav-janu, line 647: this is probably a problem with numeric options

stanislav-janu:
Why compare values and if match value, then make global selection? This can be simplify with css selector.

dakorpar:
I'll just add that both versions (current and this one) works just fine, don't understand how current version doesn't work for @stanislav-janu, but I give thumbs up for merging since it does look simpler

stanislav-janu:
> I'll just add that both versions (current and this one) works just fine, don't understand how current version doesn't work for @stanislav-janu, but I give thumbs up for merging since it does look simpler

Try it with numeric values. Does not work. Tested in latest Chrome.

stanislav-janu:
Where is a problem with this fix? Why is not merged?

assets/datagrid.js
paveljanda: Why is that?

stanislav-janu: This is wrong because there is strict comparsion of type, but teher we need "1" == 1. And this notation is unnecessarily long. `input.find("option[value='" + valueToEdit + "']")` fixed this.

paveljanda:
Thanks! Also for the explanation. 👍

Issue #899, opened 2020-05-06
StringConfirmation with Naja - cant cancel


darkWolf-PR:
I´m using Datagrid with Naja and when adding StringConfirmation to addAction like:

```php
        $grid->addAction('delete', 'admin.delete', 'delete!')
             ->setIcon('trash')
             ->setTitle('admin.delete')
             ->setClass('btn btn-xs btn-danger ajax')
             ->setConfirmation(
                 new StringConfirmation('admin.really_delete_item', 'name')
             );
```

It´s impossible to cancel action-it always submits handle and deletes item no matter what button I click in confirm alert. Like it´s not recognizing the cancel action.

Datagrid: 6.2.17 - updated today by composer from I think 6.2.11(also didn´t work, that´s why I tried to update), manually updated scripts/css from assets folder(not using npm/etc.).
Naja: 1.7
jQuery, Bootstrap4: actual versions

Naja initialized like this(from Naja author Jan Pudil):
document.addEventListener('DOMContentLoaded', naja.initialize.bind(naja));

Other javascript/ajax like filtering, pagination etc. works as expected so I guess there shouldn´t be a problem with Naja init.

paveljanda:
Closed (#900)

PR #900, opened 2020-05-06
Dont send request when confirm cancel clicked for naja.js

MartinSzollos2016:
When you click Cancel in confirm message window, request was sent anyway. This is a problem especially for delete signals.

In Naja.js you have to call interaction event instead of before to stop request.
Doc: https://naja.js.org/#/extensions-custom


MartinSzollos2016:
Duplicate of: https://github.com/contributte/datagrid/issues/899

paveljanda:
Okey, cool, thanks

PR #901, opened 2020-05-13
Hotfix/action callback onclick type

xsuchy09:
Just onClick hint type - should be `callable[]` so as array ... PHPStan is helping here.

PR #902, opened 2020-05-14
Hotfix/action callback onclick type

xsuchy09:
Just onClick hint type - should be callable[] so as array ... PHPStan is helping here.

I hope that this second attempt will not show you any conflicts as in #901 :-D


paveljanda:
thanks

PR #905, opened 2020-05-27
assets: replace pushState to replaceState in history API

martenb:
This should fix browsers BACK button funcionality, when in some cases (like filtering), you must hit back button 3 times to go back...

Not sure, if it is necessary for nette.ajax.js too in assets/datagrid-instant-url-refresh.js


paveljanda:
Nice, thanks! I will look into it

PR #906, opened 2020-05-28
Fix - paginator option 'All' does not work

Sikipiki:
Problem is with array_key_exists(). There must be in_array().

Variable $items_per_page_list looks like this:
0 => 10
1 => 20
2 => 50
3 => "all"
Even if it is then translated into e.g Czech.


paveljanda:
Thanks!

PR #907, opened 2020-06-03
Update template.md

chaeMil:
Add not so obvious reference to $item value in custom column template

### DIFF
--- a/.docs/template.md
+++ b/.docs/template.md
@@ -45,7 +45,7 @@ Or you can define column template by defining special block(s):
 {/define}
 
 {define col-title}
-	{$item} {* displays the title value *}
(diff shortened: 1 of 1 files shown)


.docs/template.md
```diff
@@ -44,6 +44,10 @@ Or you can define column template by defining special block(s):
 	:)
 {/define}
 
+{define col-title}
+	{$item} {* displays the title value *}
```
paveljanda, line 48: Could you pls alter it to for example `{$item->title}`? Thanks!

paveljanda, line 48: Could you pls alter it to for example {$item->title}? Thanks!

paveljanda:
Thanks!

PR #911, opened 2020-06-30
ColumnStatus allow option value: null

milosjanda:
Sometime default status can be null, so it would be fine implement this.

Screenshot 2020-06-30 at 21 37 34

### DIFF
--- a/src/Column/ColumnStatus.php
+++ b/src/Column/ColumnStatus.php
@@ -13,7 +13,7 @@ use Ublaboo\DataGrid\Traits\TButtonCaret;
 use Ublaboo\DataGrid\Traits\TRenderCondition;
 
 /**
- * @method onChange(string $id, string $value)
+ * @method onChange(string $id, string|null $value)
  */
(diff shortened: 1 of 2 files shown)


paveljanda:
This is actually a BC break. Sorry, I won't merge this pull request into master now.

You can simulate that behaviour using for example a MySQL function `IFNULL` (`IFNULL(test_result, "no-result") as test_result`) in combination with:

```php
$grid->addColumnStatus('statuss', 'Status')->setOptions([
	'excused' => 'Omluven',
	'no-result' => 'Nedostavil se'
]);
```

milosjanda:
Sorry i did't see it, no problem.

paveljanda:
Thanks anyway. I hope that the script helped you a bit. :)

milosjanda:
Unfortunetly no, because i'm using Nextras/Orm and Entity. But i will change default status to any defined value (like 'no-result') in MySql instead of null, which solve it. But thank you very much for very nice and usable Grid.

PR #915, opened 2020-08-29
Sorting wont use anything but ASC/DESC strings; Closes #914

### LINKED ISSUE #914: DoctrineDataSource sorting issue allowing SQL inject
Hello,
today i checked logs of new web we are launching only to find this error:
```
[2020-08-27 16-02-26] Doctrine\ORM\Query\QueryException: [Syntax Error] line 0, col 125: Error: Expected end of string, got 'DES' in /path/to/web/vendor/doctrine/orm/lib/Doctrine/ORM/Query/QueryException.php:54 caused by Doctrine\ORM\Query\QueryException: SELECT e FROM App\Entity\Marketplace e INDEX BY e.id LEFT JOIN e.item i WHERE e.storeType IN(:types) ORDER BY e.amount DES in /path/to/web/vendor/doctrine/orm/lib/Doctrine/ORM/Query/QueryException.php:43  @  https://www.example.cz/page/with-datagrid?grid-perPage=10&grid-sort%5Bamount%5D=DES  @@  exception--2020-08-27--16-01--e6985fede0.html
```

If you check carefully, string "DES" is send from URL directly do SQL which is unthinkable.

This is potentially issue with other data sources too I did not check.

juniwalk:
This should fix the sql inject i mentioned in #914, I will try to fix in in version 5.* too.

### DIFF
--- a/src/DataGrid.php
+++ b/src/DataGrid.php
@@ -824,6 +824,10 @@ class DataGrid extends Control
 		foreach ($sort as $key => $order) {
 			unset($sort[$key]);
 
+			if ($order !== 'ASC' && $order !== 'DESC') {
+				continue;
+			}
+
 			try {
 				$column = $this->getColumn($key);
 


paveljanda (changes requested):
Thanks!! Could you please fix travis? (Instead <> use !==). Thanks a lot

juniwalk:
Alright it is changed, the `<>` should not be problem travis had different issues but I changes it either way.

paveljanda (approved):

paveljanda:
Thanks!

PR #916, opened 2020-08-29
Sorting wont use anything but ASC/DESC strings; Closes #914

### LINKED ISSUE #914: DoctrineDataSource sorting issue allowing SQL inject

juniwalk:
SQL inject fix for this older branch as it is security fix. See #914 & #915 for more information.

### DIFF
--- a/src/DataGrid.php
+++ b/src/DataGrid.php
@@ -852,6 +852,10 @@ class DataGrid extends Nette\Application\UI\Control
 		foreach ($sort as $key => $order) {
 			unset($sort[$key]);
 
+			if ($order !== 'ASC' && $order !== 'DESC') {
+				continue;
+			}
+
 			try {
 				$column = $this->getColumn($key);
 


paveljanda:
Thanks!

PR #920, opened 2020-09-22
support nextras/orm 4.0 with 3.1 backward compatibility

martenb:


martenb:
ping @paveljanda

f3l1x:
I like it. We should drop support of nextras/orm v3 in the future.

paveljanda:
Thanks

PR #925, opened 2020-10-30
Added check if filter range value is numeric for DoctrineDataSource

juniwalk:
Range filter is inherently for integer values, however when you try to fill in text instead, it will throw exception.


paveljanda:
Okey, thanks 👍

juniwalk:
Thank you!

PR #930, opened 2020-11-27
Document array datasource limitation

halaxa-roman:
Documents array datasource limitation in inline add when redrawing data, with an example how to solve it


paveljanda:
Thanks!

PR #931, opened 2020-12-02
allows PHP 8.0 on composer

dg:


paveljanda:
Thanks 👍

real-antonin:
You still can't install datagrid because

`contributte/application v0.4.1 requires php ^7.2`

PR #936, opened 2020-12-13
IDataSource::getData change return type from array to iterable

AliAmjid:
**Motivation**
When implementing IDataSource client can return objects like Symfony/Collection or Nette/ArrayList and there is no need to convert it to array


f3l1x:
Make sense to me. WDYT @paveljanda?

paveljanda:
I like it. Since iterable is here since php 7.1, we can merge it 👍

paveljanda (approved):

paveljanda:
Thanks

PR #937, opened 2020-12-18
support nextras/orm 4.0 with 3.1 backward compatibility

martenb:


src/DataSource/NextrasDataSource.php
```diff
@@ -41,6 +41,11 @@ class NextrasDataSource extends FilterableDataSource implements IDataSource
 	 */
 	private $dbalCollectionClass;
 
+	/**
+	 * @var int
+	 */
+	private $ormVersion;
```
vody105, line 47: @martenb Thank you for your contribution but I do not think it is a good idea to indroduce new property ormVersion next to dbalCollectionClass. I think you can rewrite the code and only use one of them because they serve the same purpose.

martenb: @vody105 ok, changed

f3l1x:
Tested, thanks.

PR #941, opened 2021-01-07
Fix ExportCsv blank output

ikeblaster:
Regression from a56c9d3563, which results in rows with no columns (since `getColumns` never returns `null`).


paveljanda:
Thanks!

PR #943, opened 2021-01-18
Fix type error in str_replace in PHP 8.0

Sikipiki:
In PHP 8.0 in some situation is fired TypeError: Argument 2 ($replace) must be of type array|string, int given

$grid->addAction('name', 'title', 'href')
 ->setConfirmation(
  new StringConfirmation('Delete number: %s?', 'number')
 );

Where "number" is defined as integer in database.


paveljanda:
Thx 👍

PR #949, opened 2021-01-26
Group action - disabled button fix

BigOHenry:
Fix for version 6.6.1.

Issue:
https://forum.nette.org/cs/33571-ublaboo-datagrid-group-action-disabled#p214078


paveljanda:
Nice, thanks!

Issue #950, opened 2021-01-26
GroupAction is not processed on Nette\Forms 3.1


BigOHenry:
With datagrid version 6.6.1/dev-master and Nette\Forms 3.1.0 there is no possible to submit GroupActions.

Issue is in Ublaboo\DataGrid\GroupAction\GroupActionCollection in method at line 170 - 177:

```
		$values = (array) $form->getValues();
		$values = $values['group_action'];

		if (
			($submitter->getName() === 'submit' && $submitter->isSubmittedBy())
			 && ($values->group_action === 0 || $values->group_action === null)) {
			return;
		}
```

`getValues` do not return group_action control with value.

**Possible reason:**
At release notes of Nette\Forms 3.1.0: _Form::getValues() returns only controls in scope (BC break)_

paveljanda:
Thanks for the info. I am not a fan of Nette releasing BC breaks in minor version. :/ Uff.

vodictomas:
Can you please check referenced PR, if this issue can be fixed this way?

paveljanda:
New datagrid now requires nette/forms ^3.1.3. See https://github.com/contributte/datagrid/releases/tag/v6.8.0

PR #953, opened 2021-02-11
GroupAction with nette/forms v3.1.2

vodictomas:
bug caused by nette/forms v3.1.0 #950

### DIFF
--- a/src/GroupAction/GroupActionCollection.php
+++ b/src/GroupAction/GroupActionCollection.php
@@ -167,7 +167,7 @@ class GroupActionCollection
 			return;
 		}
 
-		$values = (array) $form->getValues();
+		$values = $form->getUnsafeValues(\Nette\Utils\ArrayHash::class);
 		$values = $values['group_action'];
 
 		if (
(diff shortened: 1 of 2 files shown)


kravcik:
Also maybe change function `submitted`, there is no `onSuccess`, but `onSubmit`.

paveljanda:
@vodictomas Thanks, man! Could you please resolve the conflicts? Will be merged afterwards. 👍

vodictomas:
@paveljanda I cannot find a way to resolve the phpstan error, everything I tried is still failing... do you have any idea how to resolve this problem?

src/GroupAction/GroupActionCollection.php
```diff
@@ -167,7 +167,7 @@ public function submitted(NetteForm $form): void
 			return;
 		}
 
-		$values = (array) $form->getValues();
+		$values = $form->getUnsafeValues(null);
```
dakorpar, line 170: This would mean you have to have nette forms 3.1.2 minimum, and I believe this won't be merged because of it. You should maybe check if getUnsafeValues exists and call it, if it doesn't then call getValues instead...

paveljanda:
Thanks for your help. New datagrid now requires nette/forms ^3.1.3. See https://github.com/contributte/datagrid/releases/tag/v6.8.0

PR #958, opened 2021-03-02
fix build, add missing extensions to php in github actions

dakorpar:
Build is still failing because of a bug in phpstan-nette
https://github.com/phpstan/phpstan-nette/issues/63
When that will be fixed and hopefuly soon all checks should be green.


paveljanda:
Love it, thanks 👍

PR #959, opened 2021-03-03
Update README.md

petrparolek:
- polished readme
- fixed links iin readme

### DIFF
--- a/.docs/index.md
+++ b/.docs/index.md
@@ -11,7 +11,7 @@ Table of contents
 Download this package using composer:
 
 ```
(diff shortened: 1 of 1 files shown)


paveljanda:
Thanks for the MR, but it does not make sense to me. I want to copy that code hence I will copy the whole "pre" block and that should not contain any `$` sign..

f3l1x:
Just adding poll for the future. :-)

image

https://pehapkari.slack.com/archives/C09DZP6PR/p1614857403018900

PR #968, opened 2021-03-20
Support for sum aggregation function in Nextras DataSource

milsorm:
Support for using sum aggregation also in Nextras DataSource


paveljanda:
Cool, thanks 👍

PR #969, opened 2021-03-21
Suppress warning with new nette/forms 3.1

milsorm:


src/DataGrid.php
```diff
@@ -1491,7 +1491,7 @@ public function filterSucceeded(NetteForm $form): void
 			return;
 		}
 
-		$values = (array) $form->getValues();
+		@ $values = (array) $form->getValues();
```
MartkCz, line 1494: $form->getUnsafeValues() method is better than @

paveljanda: Neither is a solution. Can we find a solution for correctly setting fitler form scope?

jaroslavlibal: @paveljanda I agree that `getUnsafeValues()` is probably not an ideal solution, but it will (at least temporarily) ensure the original (expected) behavior of the grid, unlike the current situation which may not be correct? Or do I not understand Nette's new behavior correctly?

jaroslavlibal: Just an update based on a cursory analysis - the problem probably occurs in cases where there is a select filter on one (or more) of the hidden columns. In the current release the default (or active) filter on a hidden column leads to an incorrect data filtering, because although the warning is suppressed on production, the filter data are not passed by the getValues() method and the filter is not applied.

Therefore, I think that the replacement for getUnsafeValues() is suitable as a temporary workaround to make the library compatible with Nette 3.1 as fast as possible.

f3l1x:
Hey @milsorm. Can you change it to `getUnsafeValues` and rebase it?

f3l1x:
Solved in https://github.com/contributte/datagrid/pull/969/

s1t2e3:
@f3l1x Ahoj. Kedy tato zmena bude v stabilnej verzii??

PR #970, opened 2021-04-03
Fix doc TOC.

vanam:
Rename doc files in order to match TOC.


paveljanda:
Could you please edit the index.md instead of renaming the files? I am not sure whether it would work on https://contributte.org/packages/contributte/datagrid/data-source.html then. Thanks a lot

vanam:
Fixed.

f3l1x:
Thx

PR #971, opened 2021-04-05
Allow custom implementation of confirmation dialog

foxycode:
- Allow custom implementation of confirmation dialog
- Added CallbackConfirmation and StringConfirmation shortcuts
- StringConfirmation: allow multiple placeholders


f3l1x:

paveljanda:
Does it work? I have no time to test it. If it works, I would merge it, sure. 👍

PR #973, opened 2021-04-23
[7.x] Translator: switch from ITranslator to Translator

f3l1x:


paveljanda:
That would require also a change in `composer.json` because `nette/utils` @ `v3.0` does not suppoer Nette\Localization\Translator, only `ITranslator`.

f3l1x:
Well, not necessary, look.

```
➜ composer why nette/utils
ublaboo/datagrid       6.7.x-dev  requires  nette/utils (^3.0.1)
nette/application      v3.0.0     requires  nette/utils (^3.0)
nette/caching          v3.0.1     requires  nette/utils (^2.4 || ^3.0)
nette/component-model  v3.0.0     requires  nette/utils (^2.5 || ^3.0)
nette/database         v3.0.2     requires  nette/utils (^3.0)
nette/di               v3.0.6     requires  nette/utils (^3.1.4)
nette/finder           v2.5.2     requires  nette/utils (^2.4 || ^3.0)
nette/forms            v3.1.3     requires  nette/utils (^3.2.1)
nette/http             v3.1.0     requires  nette/utils (^3.1)
nette/php-generator    v3.3.3     requires  nette/utils (^2.4.2 || ^3.0)
nette/robot-loader     v3.2.0     requires  nette/utils (^3.0)
nette/routing          v3.0.0     requires  nette/utils (^3.0)
nette/schema           v1.1.0     requires  nette/utils (^3.1.4)
nextras/orm            v3.1.0     requires  nette/utils (~2.5 || ~3.0@rc)
```

```
composer why nette/forms
ublaboo/datagrid  6.7.x-dev  requires  nette/forms (^3.1.3)
```

----

We require nette/utils 3.2.1 by default. We can raise it.

paveljanda:
In some scenarios, it's OK. But our `composer.json` is saying "we support nette 3.0". That is not  true.

As I am thinking about it more and more...

1) It is a BC break, we would have to release a new major version
2) I don't really understand nette/utils in current state:
    - if false? https://github.com/nette/utils/blob/master/src/compatibility.php#L25
3) As far as there is ITranslator supported, I would stick to it

Let's leave this MR open until Nette drops ITranslator support or until we release a new major version.

f3l1x:
Agreed. I have some weird issue with `ITranslator` in our app and PHP 8 + JIT. We're running on this branch for now.

radimvaculik:
@f3l1x Please, rebase, and merge it into `fx/bootstrap5` or `feature/php8` branch 🙏

PR #975, opened 2021-05-20
Convert primaryKey value to int if it is numeric

juniwalk:
I have issue with inline edit not working when using Doctrine Collection, I tracked issue to the type of the value when filtering for one row.

I don't know if I should leave it like this or treat the issue globally and cast all numerics to ints? What do you think?


paveljanda:
👍 Thanks

paveljanda:
I would not cast all numerics to int as it is a potential BC break..

paveljanda:
Let's try this one and observe how many people will be sad about this change.

juniwalk:
Alright, thank you!

juniwalk:
Hi, any plans to bump the version? I am finishing project that needs this fix and I'd like to have it on stable version :)

paveljanda:
👍 https://github.com/contributte/datagrid/releases/tag/v6.9.0

PR #980, opened 2021-06-29
Improved group action submit button selector

juniwalk:
Hi,
I was getting group actions to work in my custom BS4 template, however I wanted the button to have a icon, so I used `<button>` which then did not work with button enabling.

This modification should support both buttons and inputs of type submit while also containing the disabling/enabling to just group actions.


f3l1x:
LGTM

paveljanda:
Thanks 👍

juniwalk:
@paveljanda Sure, I should first fix the package.json, right?

PR #981, opened 2021-07-05
Fix for setting default sort on filter reset

jerrx:


f3l1x:
LGTM

paveljanda:
This is a BC break and would cause unwanted behaviour. When you click the "reset filter" button, sort should be reset.

PR #997, opened 2021-11-16
Fixed ignoring input content in filter

TonnyJe:
With new Nette forms there is an error, which causes that values from filter are ignored.
This PR should provide fix.

### DIFF
--- a/src/DataGrid.php
+++ b/src/DataGrid.php
@@ -1491,7 +1491,7 @@ class DataGrid extends Control
 			return;
 		}
 
-		$values = (array) $form->getValues();
+		$values = $form->getUnsafeValues('array');
 
 		if ($this->getPresenterInstance()->isAjax()) {
 			if (isset($form['group_action']['submit']) && $form['group_action']['submit']->isSubmittedBy()) {


f3l1x:
Thank you

elring:
Hi, I don't know if I should respond to closed PR, but I'll try :)

Are you sure this is right solution? For me it throws error because you pass the parameter returnType as 'array', so the result is array. But at line 1592 of DataGrid.php checks $values['filter'] to be ArrayHash (`if (!$values instanceof ArrayHash)`, $values is now only $values['filter']) or it throws exception.

So correct line should be (similar to previous getValues):
`$values = (array) $form->getUnsafeValues(null);`
ie. return ArrayHash, make it array (but array items are still ArrayHash).

I can make PR if you want.

src/DataGrid.php
```diff
@@ -1491,7 +1491,7 @@ public function filterSucceeded(NetteForm $form): void
 			return;
 		}
 
-		$values = (array) $form->getValues();
+		$values = $form->getUnsafeValues('array');
```
petrparolek, line 1494: Attention!

BC break in version 6.9.3 according this PR

```
		$inlineAdd->onSubmit[] = function (ArrayHash $values): void {
			bdump($values);
			$this->flashMessage('Record was added!', 'success');
			$this->redrawControl('flashes');
		};
```

Previously typeint was `Nette\Utils\ArrayHash` and typehint is `array` now.

PR #1001, opened 2021-12-03
Fix filter throws exception ($values are not array of ArrayAssoc)

elring:
Fix for fix #997 - $values should be array of ArrayHash.

DataGrid.php checks $values['filter'] to be ArrayHash or it throws exception.

$values before PR:
obrazek

$values after PR:
obrazek

### DIFF
--- a/src/DataGrid.php
+++ b/src/DataGrid.php
@@ -1491,7 +1491,7 @@ class DataGrid extends Control
 			return;
 		}
 
-		$values = $form->getUnsafeValues('array');
+		$values = (array) $form->getUnsafeValues(null);
 
 		if ($this->getPresenterInstance()->isAjax()) {
 			if (isset($form['group_action']['submit']) && $form['group_action']['submit']->isSubmittedBy()) {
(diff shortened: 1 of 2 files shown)


paveljanda:
Would it be possible to add it to tests? Thanks! 👍

elring:
Wow. I really don't know. No tests exist for filter. I've only once modified some test, never created new :-(
I will try something...

paveljanda:
If you have any trouble with it, just tell me, I will try to prepare something. But if you could do it, that would be awesome. 👍 Thanks, appreciated.

elring:
@paveljanda I've made some very simple FilterTest. No advanced data filtering checking this time. Only checks if filterSucceed function "succeeds" (no exception is thrown). 

Before my change datagrid fix the test fails (UnexpectedValueException), after my fix the test is ok.

I don't really know how to make tests so I've only copied the DataGridTest.phpt and make few changes to call filterSucceed method.

milsorm:
Any plans to merge fix with devmaster? Because this is really pissing, without #997 it is necessary to patch it everytime when updated :-( ?

PR #1005, opened 2021-12-21
add new snippet - toolbarButtons

Ivorius:
With this new snippet is possible add reset button up of grid

Something like this
```
		$grid->onAnchor[] = function () use ($grid): void {
			$grid->addToolbarButton($grid->getName() . '-resetFilter!', 'reset filtru')
				->setRendererOnCondition(
					fn () => Html::el('span'),
					fn () => !$grid->isFilterActive()
				)
				->setClass('ajax btn btn-warning btn-sm reset-filter');
		};

		$grid->onFiltersAssembled[] = function () use ($grid): void {
			$grid->redrawControl('toolbarButtons');
		};
```

### DIFF
--- a/src/templates/datagrid.latte
+++ b/src/templates/datagrid.latte
@@ -93,9 +93,11 @@
 							{/if}
 
 							<div class="datagrid-toolbar" n:if="$control->canHideColumns() || $inlineAdd || $exports || $toolbarButtons">
+								{snippet toolbarButtons}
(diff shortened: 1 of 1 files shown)


f3l1x:
Thank you

PS-CVIS:
The toolbarButtons snippets broke layout with exports and settings buttons. Now toolbar buttons are not in row with exports and settings buttons.

PR #1006, opened 2021-12-21
validation scope filter

Ivorius:
It fix problem with filtration if the datagrid is with inlineEdit or inlineAdd with selectboxes. Works for `$grid->setAutoSubmit(false);`


radimvaculik:
@Ivorius Looks good to me. Could you rebase the branch, please?

PR #1012, opened 2022-01-14
fix FilterSelect applying

nidjo17:


nidjo17:
This code fix issue with FilterSelect in case when datasource is array. Filter is currently not working as expected.

src/DataSource/ArrayDataSource.php
```diff
@@ -192,6 +193,11 @@ protected function applyFilter($row, Filter $filter)
 
 			$condition = $filter->getCondition();
 
+			if ($filter instanceof FilterSelect) {
+				$key = (string) key($condition);
+				return $row[$key]  === $condition[$key];
+			}
+
```
paveljanda, line 200: Thanks for the PR. I have 2 things:

1) Pls, could you add some tests? Thanks, appreciated.
2) That solution would work also for the code below. It is (in a way) duplicit. Also, why are you using different naming scheme? (column vs key)

You can do for example that:

```php
foreach ($condition as $column => $value) {
	$value = (string) $value;
	$rowVal = (string) $row[$column];

	if ($filter instanceof FilterSelect) {
		return $rowVal === $value;
	}

	if ($filter instanceof FilterText && $filter->isExactSearch()) {
		return $rowVal === $value;
	}

	$words = $filter instanceof FilterText && $filter->hasSplitWordsSearch() === false ? [$value] : explode(' ', $value);
	$row_value = strtolower(Strings::toAscii((string) $row[$column]));

	foreach ($words as $word) {
		if (strpos($row_value, strtolower(Strings::toAscii($word))) !== false) {
			return $row;
		}
	}
}
```

or:

```php
$column = key($condition);
$value = (string) $condition[$column];
$rowVal = (string) $row[$column];

if ($filter instanceof FilterSelect) {
	return $rowVal === $value;
}

if ($filter instanceof FilterText && $filter->isExactSearch()) {
	return $rowVal === $value;
}

$words = $filter instanceof FilterText && $filter->hasSplitWordsSearch() === false ? [$value] : explode(' ', $value);
$row_value = strtolower(Strings::toAscii((string) $row[$column]));

foreach ($words as $word) {
	if (strpos($row_value, strtolower(Strings::toAscii($word))) !== false) {
		return $row;
	}
}
```

paveljanda (changes requested):

src/DataSource/ArrayDataSource.php
nidjo17: Fixed and pushed. Is it ok now?

PR #1016, opened 2022-02-12
docs: fix assets for non-npm users 

elden43:
to use valid html template that matches current composer version to avoid severe problems (disabled group-action buttons...)


f3l1x:
Thanks.

PR #1018, opened 2022-02-14
Fix for PHP 8.1 "Implicit conversion from float to int loses precision"

elring:
Float value in number column with PHP 8.1 throws deprecated error - "Implicit conversion from float 123.56 to int loses precision". 

This PR fixes this by explicit type casting floats to int (for use as array index).


paveljanda:
Could you please fix the CI errors? Thanks

elring:
Done, codesniffer happy :)

paveljanda:
Nice, thanks!

PR #1020, opened 2022-02-27
GroupActionMultiselect fix

jan-stanek:
I had an issue that multiselect for multiselect group action was permanently hidden, it works after this change.

I also changed multiselect button class to `btn-light` which is more similar to BS3 `btn-default` and standard select.


f3l1x:
Please, rebase it if it is still relevant.

PR #1028, opened 2022-04-06
Allow zero values in range filter for Nextras Data Source

demmmmios:
allow zero values in range filter for Nextras Data Source. Taken from Doctrine datasource


f3l1x:
Thank you

paveljanda:
@demmmmios Could you please provide some tests? Thanks.

PR #1027, opened 2022-04-05
Compatibility with Latte 2.11 & 3

dg:

PR #1029, opened 2022-04-11
fix: snippet id

TomPavelec:


paveljanda:
I guess this is a duplicate of https://github.com/contributte/datagrid/pull/1027

Closing for now as it is already prepared in mentioned PR + Latte 3 is still in Beta. If I am mistaken and it fixes something else (I can't see any description here), please let me know.

THanks

TomPavelec:
Yes it look like same issue.
But it occurs also with latte v 2.11.1

ikeblaster:
@paveljanda If I am not overlooking something, then this fix was later removed from #1027. No idea why. Screenshot of the original fix:

image


As OP said, it's also needed for 2.11.1, which is current stable version and should be compatible with ublaboo/datagrid. Yet it throws 
> User Deprecated
> The expression 'item-{$row->getId()}' should be put in double quotes.

---

Funnily enough - latte/latte is not even a requirement for ublaboo/datagrid, not even through nette/application (even though `Nette\Bridges\ApplicationLatte\Template` uses `Latte` package, so probably not "our" problem).

PR #1031, opened 2022-05-19
Update the filterMultiSelect initialization

jaroslavlibal:
@radimvaculik prepared a workaround for filterMultiselect that ensured that the "form-select" and "form-select-sm" classes on selects are replaced with the "form-control" and "form-control-sm" ones. This ensures that the "duplicate borders" of the bootstrap-select disappears.

However, in the original implementation this was registered to the success event of the "datagrid.fitlerMultiSelect" extension registration. So if I understand the implementation correctly, the class replacement is only done when some naja/nette.ajax.js request succeeds. That's why the "duplicate borders" remained for me when the page was loaded without any interaction.

So I moved the replacement of these classes to the datagridFilterMultiSelect() function, which is called both on success event and on page load.

I also replaced the slectpicker('refresh') call with a selecpicker('destroy') and his new initialization pair (workaround for a bug in bootstrap-select - https://github.com/snapappointments/bootstrap-select/issues/2738 ).

### DIFF
--- a/assets/datagrid.js
+++ b/assets/datagrid.js
@@ -112,7 +112,7 @@ if (typeof naja !== "undefined") {
 }
 
 
-var datagridFitlerMultiSelect, datagridGroupActionMultiSelect, datagridShiftGroupSelection, datagridSortable, datagridSortableTree, getEventDomPath,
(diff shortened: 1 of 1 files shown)


paveljanda:
Could you please provide a short description? Thx

jaroslavlibal:
I've just updated the initial description.

paveljanda:
Thanks!

PR #1034, opened 2022-06-10
Remove group action input requirement double translation

elring:
The group action with text input (addGroupTextAction) makes an extra translation of message about required values. 
But the form has been already translated (function createComponentFilter in Datagrid.php calls setTranslator) so the final message is wrong.

This PR simply removes the extra translation.


paveljanda:
Thank.

Oh, could you send another PR with a test? Thanks!!

elring:
Sure, will try :)

PR #1037, opened 2022-07-16
 Effective DataModel when used with Rest API based datasource

VojtechBuba:
When Datagrid is used with common database like data sources, it is common, that receiving total count for paging cost extra one query. When working with REST API datasource it is best practice that response for getting resource collection contains information about total count.

 For the purposes of creating query only offset and limit is necessary. The total count is necessary for two reasons. First is to show UI component with list of pages. Second to check if requested page number is not higher that collection maximum.

 The problem with current model is, that the order in witch get total count is received cost two calls on Rest API endpoint instead of one. I suggest to move call of getCount() method after getting data. When working with DB, it does not matter if the getCount() call is sooner or later with performance in mind.

 If this kind of change is too big BC break, another possibility would be to create interface for data model and enable change of DataModel. One will be more strict in terms of pagination overflow and another less.

### DIFF
--- a/src/DataModel.php
+++ b/src/DataModel.php
@@ -139,7 +139,6 @@ final class DataModel
 		 */
 		if ($paginatorComponent !== null) {
 			$paginator = $paginatorComponent->getPaginator();
-			$paginator->setItemCount($this->dataSource->getCount());
 
 			$this->dataSource->sort($sorting)->limit(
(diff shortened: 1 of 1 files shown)


VojtechBuba:
Closing in favor of https://github.com/contributte/datagrid/pull/1038

PR #1038, opened 2022-07-17
Introducing interface for DataModel

VojtechBuba:
When Datagrid is used with common database like data sources, it is common, that receiving total count for paging cost extra one query. When working with REST API datasource it is best practice that response for getting resource collection contains information about total count.

For the purposes of creating query only offset and limit is necessary. The total count is necessary for two reasons. First is to show UI component with list of pages. Second to check if requested page number is not higher that collection maximum.

The problem with current model is, that the order in witch get total count is received cost two calls on Rest API endpoint instead of one. I suggest to move call of getCount() method after getting data. When working with DB, it does not matter if the getCount() call is sooner or later with performance in mind.

I have already created PR https://github.com/contributte/datagrid/pull/1037 but i`ve change my mind and created this PR.

DataModel now implements interface and Datagrid use interface instead of specific implementation. For purposes of REST based datasources i have created new DataModel where order of getCount method call is changed. 

If somebody wants to change datamodel, it can be done with new method `setDataModel`.

### DIFF
--- a/src/DataGrid.php
+++ b/src/DataGrid.php
@@ -36,10 +36,12 @@ use Ublaboo\DataGrid\Column\ItemDetail;
 use Ublaboo\DataGrid\Column\MultiAction;
 use Ublaboo\DataGrid\Components\DataGridPaginator\DataGridPaginator;
 use Ublaboo\DataGrid\DataSource\IDataSource;
+use Ublaboo\DataGrid\Exception\CannotChangeDataModelException;
 use Ublaboo\DataGrid\Exception\DataGridColumnNotFoundException;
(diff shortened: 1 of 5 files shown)


f3l1x:
Tohle vypada dobre, jen je potreba to prehodit na `next` vetve.

f3l1x:
Please, rebase it if it is still relevant.

f3l1x:
@radimvaculik I like RestModelDatasource etc. But this PR includes more features, WDYT?

f3l1x:
ping @radimvaculik

radimvaculik:
Closed in favor #1213 .

PR #1047, opened 2022-09-21
Added option of setting query hints

juniwalk:
I found myself in need of setting query hints in DoctrineDataSource. I implemented this functionality as I thought it would be good to use.

### DIFF
--- a/src/DataSource/DoctrineDataSource.php
+++ b/src/DataSource/DoctrineDataSource.php
@@ -57,6 +57,11 @@ class DoctrineDataSource extends FilterableDataSource implements IDataSource, IA
 	 */
 	protected $placeholder;
 
+	/**
+	 * @var array<string, mixed>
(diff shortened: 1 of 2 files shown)


paveljanda:
@juniwalk Could you pls fix phpstan errors?

@f3l1x Could you please check it? I am not using doctrine myself so it would be better if you review it. Thanks 👍

juniwalk:
@paveljanda already working on it.

juniwalk:
@paveljanda @f3l1x Hello, anything new? :) For review all you need is to check [this method](https://github.com/doctrine/orm/blob/2.13.x/lib/Doctrine/ORM/AbstractQuery.php#L1031).

f3l1x (approved):

f3l1x:
It's ok. Thanks @juniwalk

PR #1049, opened 2022-10-12
Fixing custom classes on checkbox in inline edit

Spilky:
- `\Nette\Forms\Controls\Checkbox::getControl` returns label part and control part like one `\Nette\Utils\Html` and then `$control->getControl()->getAttribute('class')` is always null (because it takes class from label) and custom defined classes are always overwritten by `$control->setAttribute('class', 'form-control input-sm form-control-sm');`
- fixed by using `\Nette\Forms\Controls\BaseControl::getControlPrototype` which always returns only control part without label part

### DIFF
--- a/src/InlineEdit/InlineEdit.php
+++ b/src/InlineEdit/InlineEdit.php
@@ -190,7 +190,7 @@ class InlineEdit
 					break;
 
 				default:
-					if ($control->getControl()->getAttribute('class') === null) {
+					if ($control->getControlPrototype()->getAttribute('class') === null) {
 						$control->setAttribute('class', 'form-control form-control-sm');
 					}
 


MrVoltz:
I am having the same issue, I think that in addition to your fix, there should be a condition for TextBase and SelectBox.

radimvaculik:
Good for me! 👍 @paveljanda / @f3l1x Please merge it.

f3l1x:
Please rebase your PR.

Spilky:
@f3l1x done. There is error for PHPStan and it is not related to my PR:

```
Error: Ignored error pattern #^Parameter \#1 \$x of method Doctrine\\ORM\\Query\\Expr\:\:like\(\) expects string, Doctrine\\ORM\\Query\\Expr\\Func given\.$# in path /home/runner/work/datagrid/datagrid/src/DataSource/DoctrineDataSource.php was not matched in reported errors.
```

radimvaculik:
@Spilky Fixed @ https://github.com/contributte/datagrid/pull/1158

Spilky:
@radimvaculik @f3l1x rebased again. All checks are successful. Please merge it.

PR #1054, opened 2022-12-14
Cast datagrid-editable-value to string

Rixafy:
There was a problem, when someone uses `setEditableValueCallback` on a row, and returns some string, but this string is for instance `123`, javascript casts that attribute to integer, and when you click away, callback is called because integer != string - datagrid thinks the value has changed.


radimvaculik:
Good for me! 👍  @f3l1x / @paveljanda Could you merge it, please?

paveljanda (approved):

paveljanda:
Thanks

PR #1055, opened 2022-12-14
Fix repeating of persistent parameter in url

Rixafy:
When loading clean page with datagrid and persistent parameter included in router url, datagrid js always updated url and added redundant persistent parameter, for isntance:

url where example.com is persistent parameter `domainName` => `http://localhost/domain/example.com/cron-task`
would turn into `http://localhost/domain/example.com/cron-task?domainName=example.com`
because datagrid js has no way of knowing that there is already that persistent parameter.

In that case, my solution checks for `/example.com` in url, and if it's there, it skips that parameter while building url query, I think it's good enough when there is persistent parameter standalone somewhere in url.

### DIFF
--- a/assets/datagrid.js
+++ b/assets/datagrid.js
@@ -354,6 +354,9 @@ for(var p in obj) {
 	if (obj.hasOwnProperty(p)) {
 		var k = prefix ? prefix + "[" + p + "]" : p, v = obj[p];
 		if (v !== null && v !== "") {
+			if (window.location.pathname.includes('/' + v)) {
+				continue;
+			}
 			if (typeof v == "object") {
 				var r = window.datagridSerializeUrl(v, k);
 					if (r) {


radimvaculik:
@Rixafy Have you tested with multiple parameters with same value, but different key? For example `domainName1=example.com&domainName2=example.com`

Rixafy:
@radimvaculik I cannot replicate that, when I try to accept $test, in actionDefault(string $test), and then I add ?test=example.com in url, datagrid will ignore that test parameter for some reason, and it will apply only domainName persistent parameter and grid-* parameters.

Anyway, I improved the checking if this happened, after every check it will remove that `/example.com` from checked url, and next parameter will be checked on modified url string, this will work when persistent parameter will be placed before the get parameters in that map, but I have no way of testing if it's the case, since it ignored my `?test=example.com` completely and it modified url without that parameter, so when I reloaded the page, I got 404.

f3l1x:
Please rebase it @Rixafy, if it's still relevant.

PR #1065, opened 2023-01-29
Update InlineEdit.php

emololftw:
Error: Typed property Ublaboo\DataGrid\InlineEdit\InlineEdit::$itemID must not be accessed before initialization

in main template.latte on [L#248](https://github.com/emololftw/datagrid/blob/next/src/templates/datagrid.latte#L248) file is variable check via:

{if $inlineEdit && $inlineEdit->getItemId() == $row->getId()}


radimvaculik:
Looks good to me 👍

PR #1066, opened 2023-02-03
Fix for searching id over array data source

radimSekera:
When condition for filterOne looks like `$condition = ['id' => '25661']; and object structure is ['id' => int, ...] item detail is not working correctly.

Thank you for your time

### DIFF
--- a/src/DataSource/ArrayDataSource.php
+++ b/src/DataSource/ArrayDataSource.php
@@ -93,10 +93,7 @@ class ArrayDataSource implements IDataSource
 	{
 		foreach ($this->data as $item) {
 			foreach ($condition as $key => $value) {
-				if ($key === 'id' && is_string($value)) {
-					$value = intval($value);
-				}
-				if ($item[$key] === $value) {
+				if ($item[$key] == $value) {
 					$this->setData([$item]);
 
 					return $this;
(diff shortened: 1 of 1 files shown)


radimvaculik:
@radimSekera Please, add some tests.

radimvaculik:
You assume that the key will be called `id`. But what if his name is different? For example `age` and is also `int`?

src/DataSource/ArrayDataSource.php
```diff
@@ -93,6 +93,9 @@ public function filterOne(array $condition): IDataSource
 	{
 		foreach ($this->data as $item) {
 			foreach ($condition as $key => $value) {
+				if ($key === 'id' && is_string($value)) {
+					$value = intval($value);
+				}
 				if ($item[$key] === $value) {
```
radimvaculik, line 99:
```suggestion
				if ($item[$key] == $value) {
```

What about to change to `==` to ignore types?

radimSekera: Yes i thought of that too, its better idea a should work better. Guess there is no chance of getting like `value = null` and having $item[0] for match. 
Another thing that went over my head is regex check for only numbers in value then convert to to int.

But ignoring types is easiest stuff to do. Thank you

radimvaculik: Are you going to change the pull request?

radimSekera: Pull request updated.

radimvaculik:
@radimSekera Thanks. Great. Could you introduce some more tests, please?

radimvaculik:
@radimSekera ping,

Issue #1067, opened 2023-02-09
Rename package name from `Ublaboo` to `Contributte`


radimvaculik:
It's obvious from the title.

```php
- namespace Ublaboo\DataGrid;
+ namespace Contributte\Datagrid;
```

radimvaculik:
@paveljanda @f3l1x Could you please give your opinion? Thanks.

paveljanda:
Yep, I agree, you can send a PR 👍

radimvaculik:
@lilianalillyy Could you prepare a PR on `next` branch with `compatibility.php` file? 🙏

lilianalillyy:
Sure!

lilianalillyy:
@radimvaculik Sorry it took so long but PR is opened now :)

PR #1075, opened 2023-04-15
ublaboo -> contributte (#1067)

lilianalillyy:
Renames namespace from `Ublaboo\DataGrid` to `Contributte\Datagrid`  (issue #1067) - package name not yet changed. 

Script that generates compatibility.php: https://gist.github.com/lilianalillyy/ee8de66b416fd1d680c58eba3f641d15


radimvaculik:
@lilianalillyy Thanks for PR! 🎉 Could you make CI green, please?

lilianalillyy:
This should be everything, the only things that have been left intentionally

1. `ublaboo/datagrid` library name (Composer) and `ublaboo-datagrid` (NPM) - This I'm leaving to @f3l1x / @paveljanda, who have to publish new packages (imo, the current packages should be deprecated though not removed completely, with a deprecation notice to use `contributte/datagrid`/`@contributte/datagrid` instead (both managers support this, even with the ability of referencing which package should be used as replacement). (Affected files: .docs/index,filters,assets.md, .docs/assets.md, README.md, package.json, composer.json)
2. CSS/JS assets - those are both subject to change already (rewrite)

As for translation keys, they are now prefixed with `contributte_`, rest stays the same. Before, they used the `ublaboo_` prefix, which will still function and will be automatically transformed to the contributte_ equivalent (a E_USER_DEPRECATED error will be triggered for the ublaboo_ keys). Implementation of that can be seen in SimpleTranslator.php.

radimvaculik:
@lilianalillyy Could you make CI green, please?

lilianalillyy:
`PHP Fatal error:  Declaration of Contributte\Datagrid\Filter\SubmitButton::getControl(object|string|null $caption = null): Nette\Utils\Html must be compatible with Nette\Forms\Controls\Button::getControl($caption = null): Nette\Utils\Html in /home/runner/work/datagrid/datagrid/src/Filter/SubmitButton.php on line 34`

It doesn't make sense to me, for reference:
```php
<?php
// ...
use Nette\Utils\Html;

// ...

public function getControl(string|object|null $caption = null): Html
{
 // ...
}
```

radimvaculik:
Does it fail on your local, when `make cs` ?

lilianalillyy:
nope, I only had an issue with the alphabetical sorting (which I've now fixed), this error didn't occur

radimvaculik:
I just clone your repo, run `make` and `make cs` and got the error. Running on PHP 8.0 locally.

lilianalillyy:
Removing vendor and reinstalling dependencies caused the same issue, however I fail to see how this PR affected that

radimvaculik:
Is `compatibility.php` included in phpcs/phpstan paths somehow?

lilianalillyy:
it's included in the composer autoload, however I doubt it has any impact on this as the error is regarding a Contributte namespace, and looking closely I see the issue, just not sure how it passed the CI before

lilianalillyy:
It shouldn't work in the first place, as the signature of the parent class is in fact different

radimvaculik:
But if you clone original repo, run `make` and `make cs`, it works like a charm 🤷‍♂️

lilianalillyy:
i believe it's some sort of a bug: repro:
cs doesn't work on my next branch -> checkout to d7bbdacb738565bf0adef37059b179c33104fa0f (last commit of contributte's next branch -> make cs works -> checkout to e7b9b5e9907489512b6f35747a5208293d5b8b6e (last commit of my next branch) -> make cs works now

lilianalillyy:
I have no idea what causes this, if someone could take a look it'd be great?

lilianalillyy:
It's already shared in a gist above ^

radimvaculik:
@lilianalillyy Let's go with for now.

```
-       public function getControl(string|object|null $caption = null): Html
+       /** @inheritDoc */
+       public function getControl($caption = null): Html
```

lilianalillyy:
sounds good, let me update the code

src/Localization/SimpleTranslator.php
```diff
@@ -1,44 +1,59 @@
...
 	{
+		// BC support for 'ublaboo' translations
+		$oldPrefix = 'ublaboo_';
+		$newPrefix = 'contributte_';
+		foreach ($dictionary as $key => $value) {
+			if (str_starts_with($key, $oldPrefix)) {
+				$newKey = $newPrefix . substr($key, strlen($oldPrefix));
+				// Only change the keys that are in the default $dictionary (other keys are left as-is)
+				if (array_key_exists($newKey, $this->dictionary)) {
+					trigger_error(`Translation key '$key' is deprecated, please use '$newKey' instead`, E_USER_DEPRECATED);
```
radimvaculik, line 50: @lilianalillyy Use `sprintf`

radimvaculik (changes requested):

radimvaculik:
@f3l1x @paveljanda All good to me. Could you check PR and eventually merge it? Thanks guys.

radimvaculik (approved):

radimvaculik:
@lilianalillyy Please, rebase 🙈

lilianalillyy:
Will get to it later today!

lilianalillyy:
@radimvaculik rebased! sorry it took so long, i was out of town (Brno :D)

radimvaculik:
@lilianalillyy Thanks for contributing 👍

PR #1077, opened 2023-04-20
Allow nette/utils:4.0

radimvaculik:


foxycode:
@radimvaculik Seems working.

PR #1080, opened 2023-04-21
Return value of BackedEnum when reading Doctrine entity property.

juniwalk:


juniwalk:
Should I add this elsewhere in Row class?

radimvaculik:
@juniwalk Thanks for PR. Could you introduce some tests as well, please?

juniwalk:
@radimvaculik I added needed test, however it is php8.1 feature, Should I mark it somehow to test only for php8.1?

juniwalk:
Well I fixed the test but that codesniffer error I can't find it and the output doesn't tell you where the issue is.

radimvaculik:
Could you point it to `next` branch, please?

juniwalk:
image

It should be pointed into `next`

PR #1081, opened 2023-04-22
Return value of BackedEnum when reading Doctrine entity property.

juniwalk:
See #1080

### DIFF
--- a/src/Row.php
+++ b/src/Row.php
@@ -216,6 +216,10 @@ class Row
 			return (string) $value;
 		}
 
+		if (interface_exists(\BackedEnum::class) && $value instanceof \BackedEnum) {
+			return $value->value;
+		}
+
 		return $value;
 	}
 
--- a/tests/Cases/EnumTest.phpt
+++ b/tests/Cases/EnumTest.phpt
@@ -1,0 +1,42 @@
+<?php declare(strict_types = 1);
+
+namespace Ublaboo\DataGrid\Tests\Cases;
+
+/**
+ * @phpVersion >= 8.1
+ */
+
+require __DIR__ . '/../bootstrap.php';
+require __DIR__ . '/../Files/TestingDataGridFactory.php';
+
+use Tester\Assert;
+use Tester\TestCase;
+use Ublaboo\DataGrid\DataGrid;
+use Ublaboo\DataGrid\Row;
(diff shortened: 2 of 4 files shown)


juniwalk:
@radimvaculik Alright, this should be much better, I closed #1080.

juniwalk:
@radimvaculik ping

radimvaculik:
@juniwalk How does it work on PHP 8.0, where BackedEnum is not implemented?

juniwalk:
@radimvaculik I think class is not autoloaded so it just ignores it. That is why php8.0 did not fail.

src/Row.php
```diff
@@ -216,6 +217,10 @@ public function getDoctrineEntityProperty(mixed $item, mixed $key): mixed
 			return (string) $value;
 		}
 
+		if (is_object($value) && $value instanceof BackedEnum) {
```
radimvaculik, line 220:
```suggestion
		if (interface_exists(\BackedEnum::class) && $value instanceof BackedEnum) {
```

radimvaculik: @juniwalk I prefer something like this.

juniwalk: No problem, I will add it.

radimvaculik: Why do you check it, if `$value` is object with `is_object` ?

juniwalk: Oh I wish I knew :D I usually don't do that.

juniwalk:
@radimvaculik Alright, like that?

radimvaculik:
Good to me. But I have zero experience with Doctrine. Can anyone else have a look to check the PR?

juniwalk:
@radimvaculik This does not affect Doctrine per se. It just changes the way DataGrind handles Enums returned from Doctrine. Without this DataGrid would not render Status row as it was unable to handle BackedEnum instance.

juniwalk:
Actually if you had for example ArraySource with Enum instance it would see the same issue. DataGrid cannot handle them, mainly in StatusColumn.

radimvaculik:
Good to me.

juniwalk:
@radimvaculik I am thinking if I should add this for other sources as well, besides ArraySource I don't know which one would be able to return instance of Enum.

PR #1086, opened 2023-07-02
Add support for `nette/utils` version `4.0.0`

harmim:
This PR is adding support for `nette/utils` version `4.0.0`.

I went through all usages of `nette/utils` in the repository. It seems to be safe.

I did not know how to get rid of that PHPStan warnings related to the `\Stringable` interface which is used as a type hint in `nette/utils` other than ignore these warnings. We can not use `\Stringable` since we want to support PHP versions `<8`. Is there a better solution?

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -38,7 +38,7 @@
 		"contributte/application": "^0.5.0",
 		"nette/di": "^3.0.0",
 		"nette/forms": "^3.1.3",
-		"nette/utils": "^3.0.1",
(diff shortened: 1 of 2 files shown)


paveljanda:
I am OK with that 👍

radimvaculik:
Looks good to me too 👍

radimvaculik (approved):

dg:
Please, tag it 🙏🏼

PR #1091, opened 2023-07-18
Fix timeout vs. number utils.ts

MartinKokesCz:
Fix for this error during the build
[tsl] ERROR in /node_modules/ublaboo-datagrid/assets/utils.ts(130,3)
      TS2322: Type 'Timeout' is not assignable to type 'number'.

### DIFF
--- a/assets/utils.ts
+++ b/assets/utils.ts
@@ -120,7 +120,7 @@ export function debounce<TArgs, TFun extends (...args: TArgs[]) => unknown | Pro
 	fn: TFun,
 	slowdown: number = 200
 ): (...args: TArgs[]) => void {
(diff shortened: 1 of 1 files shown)


radimvaculik:
@lilianalillyy Could you check it, please?

lilianalillyy:
lgtm, this is due to the node.js environment having a different return value for setTimeout than the browser environment, which at runtime doesn't matter (in this case at least) but typescript will scream depending on the build environment setup; this is a good fix for that, thank you!

lilianalillyy (approved):

radimvaculik:
@MartinKokesCz Thanks for contributing. @lilianalillyy Thanks for code review 💪

PR #1092, opened 2023-08-02
fix: resolution name for datagrid -> attribute instead of class

to23mas:
Fix for warning in developers console. image


f3l1x:
Awesome 👍

PR #1095, opened 2023-08-14
Request values of inline add/edit containers only when this part of form is submitted

vvoody-:
Inline add/edit form container can be mapped to object with required properties, so creation of this value object will fail when form controls are set as required but their validation is skipped by validation scope for example when grid filter is submitted.

### DIFF
--- a/src/Datagrid.php
+++ b/src/Datagrid.php
@@ -1262,8 +1262,6 @@ class Datagrid extends Control
 			return;
 		}
 
-		$values = (array) $form->getUntrustedValues(null);
-
 		if ($this->getPresenterInstance()->isAjax()) {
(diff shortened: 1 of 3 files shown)


f3l1x:
Can you please add some tests?

vvoody-:
So I rewrote this fix a bit, added test and rebased branch. I'm not sure if I did it right. Should I create new pull request or something?

f3l1x:
It's good, only one thing, can you please target this MR to master branch and not 6.x?

f3l1x:
Ok, I've changed target branch to master and there are some conflicts.

vvoody-:
Done

f3l1x:
Thank you

Issue #1101, opened 2023-10-10
Wrong display of toolbar & export buttons


elring:
After commit 55fd8b4  the toolbar look wrong:

![image](https://github.com/contributte/datagrid/assets/24586123/ba4c5c31-e184-43bf-8015-779718f1f82c)

1. The buttons are without any space between - this is because of now wrong CSS selector in datagrid.css - extra div is now needed - line 74 should be changed to:
```
.datagrid .datagrid-toolbar > div > span > a {
  margin-left: 0.5em;
}

```

2. The export buttons are on new line - this is because of the new snippet div - the export toolbar is outside of it. This can be solved either by moving the export toolbar into the new snippet or by css "hack" of the snippet div:

```
.datagrid .datagrid-toolbar > div {
  display: inline-block;
}
```

After these changes, toolbar looks better:

![image](https://github.com/contributte/datagrid/assets/24586123/46de41a5-ae68-494b-9911-40c6d79b2dfb)

Should I make PR (which way for 2)?

paveljanda:
1) Yes, pls create a new MR, thanks! 👍 
2) I am not sure I follow - what do you mean by which way for 2?

paveljanda:
Thx 👍

PR #1102, opened 2023-10-12
Fix for wrong display of toolbar & export buttons

elring:
See issue #1101
### Summary
After commit 55fd8b4 the toolbar looks wrong:

image

1. The buttons are without any space between - this is because of wrong CSS selector in datagrid.css - extra div is now needed - fixed by modifying CSS selector
2. The export buttons are on new line - this is because of the new snippet div (the export toolbar is outside of it) - fixed by CSS property display of new div set to inline-block

After these changes, toolbar looks better:

image

### DIFF
--- a/assets/datagrid.css
+++ b/assets/datagrid.css
@@ -67,12 +67,16 @@
 	display: inline-block
 }
 
-.datagrid .datagrid-toolbar > span {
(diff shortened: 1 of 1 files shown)


paveljanda:
Thanks

PR #1103, opened 2023-10-18
Allow coexistence of both outer and column filters

survik1:
Hello, we are using Datagrid on quite large project with millions of records. And we have currently reached some limitations that causes us some issues. 

We would like to create the grid overview of the users. We have currently about 800k business users. Each user can have n invoicing addresses among others (same can be applied to more records, like contacts, delivery addresses and so on). The grid displays only data from user (other entities are accessible through user detail). We need, however, to filter users based on data in their subentities. The example could be a user calling to callcenter providing their ID number (IČO) which is attribute of invoicing address. 

Datagrid allows us to use either column filters or outer filters. Keeping the columns would require to add column with ID values (probably using GROUP_CONCAT(DISTINCT)). Using outer filters would allow to add the filter we require, but it would also move all filters from columns, which isn't the desired operation as column filters are very useful. 

This PR moves rendering decision down to filter instead of keeping it on DataGrid level. The change allows to keep column filters while also add a possibility to create a detached filter that would only be added to the section over the grid. The code and its execution should be compatible with previous version, so is should be safe to deploy within next minor version. 

Example:
```php
$grid->addFilterText('ico', 'IČO')
    ->setOuterRendering()
    ->setCondition(function (Fluent $fluent, $value): void {
        if ($value) {
            $fluent->where([['invoice_address.ic = %i', $value]]);
        }
    });
```

### DIFF
--- a/src/DataGrid.php
+++ b/src/DataGrid.php
@@ -313,6 +313,7 @@ class DataGrid extends Control
 
 	/**
 	 * @var bool
+     * @deprecated 
 	 */
 	protected $outerFilterRendering = false;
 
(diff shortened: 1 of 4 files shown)


survik1:
Hello, can i get like any reaction?

survik1:
Hey guys, are you still working on this? I understand that the eventual merging may take time, but not responding is really bad sign to me.

dakorpar:
@paveljanda @f3l1x ...

survik1:
@paveljanda @f3l1x Alright guys, we are quite pilling up on fixes and updates and since this small PR can't even get any attention even in the form of "will check it out later in 2025", we will not bother bringing the other changes and just continue with the fork of the library. Thank you @dakorpar for at least your replies.

f3l1x:
Sorry @survik1. I missed this PR. Do you want to re-open it?

PR #1096, opened 2023-08-21
Datepicker input

MartinKokesCz:
Based on issue #946 I have prepared the simple HTML input type="date" for date and date range filter.

You can try it with my fork of datagrid-skeleton

https://github.com/MartinKokesCz/datagrid-skeleton/tree/next

PR #1105, opened 2023-11-17
feat: switch to native date inputs

martenb:
It looks like there is a similar MR #1096 but it's not finished so I hope that's okay.


martenb:
rebased on next branch

paveljanda:
Native datetime picker does not look like shit anymore?

martenb:
looks ok for me in chrome and firefox, ...

PR #1113, opened 2023-12-12
Update DataGrid.php

Anry7794:
Replacement of deprecated ITranslator with Translator


radimvaculik:
Already changed in `next` branch.

PR #1117, opened 2024-01-12
Changed sorting type

tmatejicek:
SORT_LOCALE_STRING - compare items as strings, based on the current locale. It uses the locale, which can be changed using setlocale()


radimvaculik:
Good to me. @f3l1x could you check it, please?

radimvaculik (approved):

PR #1128, opened 2024-05-04
Added `readonly`

radimvaculik:


paveljanda:
Yes agree.

f3l1x:
I would stay with `no readonly` code right now. If you don't mind @radimvaculik

PR #1130, opened 2024-05-04
Rename all $grid to $datagrid

radimvaculik:


petrparolek:
Why? :open_mouth:

radimvaculik:
To make it consistent across project.

paveljanda:
I like it. BUT - it's an unnecessary BC break. From this reason, I don't think we should merge it.

PR #1141, opened 2024-08-02
Update string key in data_grid_paginator.latte

dasim:


radimvaculik (changes requested):
Fix it on line 11 as well, please.

dasim:
This PR fixed both strings and you closed it as BC break, perhaps you should intergrate that one: https://github.com/contributte/datagrid/pull/1142

dakorpar:
why do we want this? Changing translation keys will screw up many working projects and doesn't really add some big benefit...  I don't see a point in forcing this,... Nice to make it consistent, but cost is to high. All projects that just upgrade will get those translations changed... It's a BC that IMHO does not make much sense to introduce.

dasim:
Unless you decide to revert what you’ve changed already, you should finish the job.

https://github.com/search?q=repo%3Acontributte%2Fdatagrid%20contributte_datagrid.&type=code
vs
https://github.com/search?q=repo%3Acontributte%2Fdatagrid+ublaboo_datagrid.&type=code

dasim:
More context on the change: https://github.com/contributte/datagrid/pull/1075 and https://github.com/contributte/datagrid/issues/1067

Issue #1146, opened 2024-09-03
Custom export takes too long


patrikvalentaa:
Hi @f3l1x, @paveljanda,

Is possible to change method getDataSource() in file Contributte\Datagrid\DataSource\NextrasDataSource.php [here](https://github.com/contributte/datagrid/blob/0105ca358f9c945874ae464b36b131d1915a5644/src/DataSource/NextrasDataSource.php#L229) visibility from protected to public?

In my case I am doing custom export, because included CSV export takes too long in case large model dependences, so I prepared custom SQL query and now i need to pass argument $userIds, which I get from filtered items from datagrid. For now there is no accessable method to get collection, here is only method getData(), but it returns Nextras Model fetchAll(), that i later iterating for get array of filtered ids. Which also takes too long. Best way for me is public visibility method getDataSource(), where I can call fetchPairs(null, 'id').

Thanx

sasule:
Seems to me like you are exporting all data and then want to filter them by filtered rows from grid.

Wouldn't be method ` $grid->addExportCsvFiltered()` a solution? 

See [docs](https://contributte.org/packages/contributte/datagrid/export.html#exportcallback).

patrikvalentaa:
Not exactly. I used to use $grid->addExportCsvFiltered(), but there is same problem with fetching all amount of data and model dependecies such as relationships, virtual properities, etc. 

[...]

PR #1147, opened 2024-09-03
getDataSource visibility protected >> public

patrikvalentaa:
Fix based on reported [issue](https://github.com/contributte/datagrid/issues/1146)


f3l1x:
@paveljanda I am OK with this. WDYT?

PR #1153, opened 2024-10-23
fix: adding multiple listeners to the same element

to23mas:
I had an issue where there was a timeout when using filters.

image

Each row on next image is added on single keypress
image

image

The reason was that the Event listener was added again with every request.


paveljanda:
Can you fix the CI, pls?

to23mas:
@paveljanda CI fixed

paveljanda (approved):

paveljanda:
Thx

PR #1164, opened 2025-02-06
Fix bugs with selectors

UltimateEvil:
Both the happy.ts and checkboxes.ts had many bugs. 
1) No initialisation after ajax relaod
2) No initialisation on page load - for counter
3) features straight up not working
- injected css via injecting the DOM
- happy checkboxes
- multirow selection
...
4) no onchange events were ever invoked so one plugin would ruin the state of the other.
5) global callbacks being registered multiple times


f3l1x:
Hi @UltimateEvil, can you please rebase it?

PR #1168, opened 2025-03-13
fix confirmation message cancel button

to23mas:
The Cancel button in pop-up messages is not functioning correctly. Instead of canceling the action or closing the pop-up, it performs the same action as the OK button.

image

image


paveljanda (approved):

paveljanda:
Thx

PR #1186, opened 2025-07-03
Fix group actions & drop HappyInputs

radimvaculik:


paveljanda:
How will it look without those pretty checkboxes? Wouldn't it look shit?

radimvaculik:
We now use native Bootstrap checkboxes, which integrate well with the rest of the UI. They still look fine.

image

See https://getbootstrap.com/docs/5.0/forms/checks-radios/

paveljanda:
Alright

PR #1180, opened 2025-06-20
Refactor session handling to use pluggable state storage

jeancz:
# Refactor session handling to use pluggable state storage

This PR introduces a pluggable state storage system that allows developers to use custom storage backends (database, cache, etc.) instead of being limited to session-only storage. This enables DataGrid state persistence across multiple devices/browsers and provides enhanced flexibility for distributed environments.

## Implementation

- **Introduced `IStateStorage` interface** with three methods: `loadState()`, `saveState()`, and `deleteState()`
- **Added two implementations:**
  - `SessionStateStorage` - maintains existing session-based behavior
  - `NoopStateStorage` - no-op implementation for when persistence isn't needed
- **Refactored `Datagrid` class:**
  - Replaced direct session access with state storage abstraction
  - Renamed session-related methods and properties to use "storage" terminology
  - Added `getStateStorage()` and `setStateStorage()` methods
  - Automatically selects appropriate storage based on `rememberState` and column hiding settings
- **Improved filter storage** by consolidating all filters under a single `_grid_filters` key instead of individual keys
- **Added deprecation notice** for `setStrictSessionFilterValues()` method
- **Updated internal naming** from "session" to "storage" throughout the codebase

## Usage Example

```php
// Use custom database storage
$grid->setStateStorage(new DatabaseStateStorage($user->getId()));
```

## Backward Compatibility

All existing functionality remains intact. The datagrid automatically uses `SessionStateStorage` when state persistence is enabled, ensuring existing applications continue to work without changes.

## Testing

Added some basic tests covering different storage scenarios including remember state.

## Notes
Documentation has not been updated yet - waiting for code review feedback before making documentation changes.

PR #1187, opened 2025-07-03
Fix: Reset column filter without deleting all grid filters

jeancz:
This change updates the handleResetColumnFilter function to correctly remove only the specific column's filter, preserving other active filters.

[See the issue](https://github.com/contributte/datagrid/pull/1180#issuecomment-3031871157)

### DIFF
--- a/.docs/README.md
+++ b/.docs/README.md
@@ -23,6 +23,9 @@
 **Export**
 - [Exports](export.md)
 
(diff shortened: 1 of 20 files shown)


radimvaculik:
LGTM 👍

@f3l1x Could you please merge this when you have a moment? Thanks!

PR #1188, opened 2025-07-04
Remove last mention about ublaboo

radimvaculik:


paveljanda:
Jjust fyi, this is a BC breaking MR

paveljanda:
Well both. If there is a BC break, you have to create a new release - 8.0.0. And describe all breaking changes. That's how semver works.

PR #1190, opened 2025-07-18
Implements the possibility to deactivate Doctrine Paginator

### LINKED ISSUE #1071: [DoctrineDataSource] Allow not to use Doctrine's Paginator
In DoctrineDataSource the Paginator from Doctrine (Doctrine\ORM\Tools\Pagination\Paginator) is used for situations when JOIN or GROUP BY is used.

I assume this is because of the "to-many" relationships - we want the number of entries on the page match the selected number. In this case, however, Doctrine runs an additional SQL query to determine the number of results, which can take quite a long time on a large databases (DISTINCT, MIN, GROUP BY...).

This (if I am only using to-one relations) might not be necessary and I would like to be able to disable the use of the external Paginator from Doctrine.

Is that a legitimate request or am I missing something, please? My idea is e.g. new public method `disableDoctrinePaginator()` that stores this request in a private variable. This is then used in the existing private method [`usePaginator()`](https://github.com/contributte/datagrid/blob/4613b83e2d10c70b73e798589d3c581d23f9f956/src/DataSource/DoctrineDataSource.php#L382).

skyfly18:
Fix #1071

### DIFF
--- a/src/DataSource/DoctrineDataSource.php
+++ b/src/DataSource/DoctrineDataSource.php
@@ -180,7 +180,6 @@ class DoctrineDataSource extends FilterableDataSource implements IDataSource, IA
 	/**
 	 * The method deactivates or activates the use of the Doctrine Paginator for JOIN and GROUP BY
 	 */
-
 	public function setUsePaginator(bool $usePaginator): void
 	{
 		$this->usePaginator = $usePaginator;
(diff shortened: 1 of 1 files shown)


src/DataSource/DoctrineDataSource.php
```diff
@@ -44,6 +44,8 @@ class DoctrineDataSource extends FilterableDataSource implements IDataSource, IA
 	/** @var array<string, mixed> */
 	protected array $hints = [];
 
+	protected bool $useDoctrinePaginator = true;
+
```
f3l1x, line 48: Please rename it to usePaginator.

PR #1191, opened 2025-07-22
V office erp mail processing

jansekera:


jansekera:
> @jansekera ?

Sorry. It had to be to Evosoft datagrid.
You may discard this branch.

PR #1194, opened 2025-08-02
Editable plugin fix

MartinKokesCz:
Here is a fix of editable plugin, which enables editing of a cell after clicking. There was an click event listener missing.


radimvaculik:
LGTM! @f3l1x

PR #1200, opened 2025-08-06
refactor: optimize filter condition handling for NextrasDataSource

emololftw:
**Refactor dataSource filtering for NextrasDataSource to use Collection function**

Refactored the filtering logic in NextrasDataSource to utilize the Collection function within filterCondition. This change improves code consistency and leverages existing Collection utilities.

**Changes:**

1. Updated filtering implementation to use Collection function inside filterCondition
2. Maintained full backward compatibility - both existing and new approaches work seamlessly

**Working cases:**
```php
$grid->addFilterText('params', 'Parameters:', 'params')
	->setCondition(function (DbalCollection $dbalCollection, $value) {
		return $dbalCollection->findBy([
			'parametersValues->parameterListValue->id' => (array) $value
		]);
	});
```

```php
$grid->addFilterText('params', 'Parameters:', 'params')
	->setCondition(function (DbalCollection $dbalCollection, $value) {
		$dbalCollection
			->getQueryBuilder()
			->joinLeft('[parameters_values_x_scrap_data] as [pvxsd]', '[scrap_data.id] = [pvxsd.scrap_data_id]')
			->joinLeft('[parameters_values] as [pv]', '[pvxsd.parameters_values_id] = [pv.id]')
			->andWhere('[pv.parameter_list_value] IN %i[]', (array) $value);
		})
	});
```

**PS: sorry, I accidentally deleted my last personal repository**

### DIFF
--- a/src/DataSource/FilterableDataSource.php
+++ b/src/DataSource/FilterableDataSource.php
@@ -32,7 +32,14 @@ abstract class FilterableDataSource
 						$value = ArrayHash::from($filter->getValue());
 					}
 
-					($filter->getConditionCallback())($this->getDataSource(), $value);
+					if ($this instanceof NextrasDataSource) {
+						$returnCheck = ($filter->getConditionCallback())($this->getDataSource(), $value);
+						if ($returnCheck) {
+							$this->dataSource = $returnCheck;
+						}
+					} else {
+						($filter->getConditionCallback())($this->getDataSource(), $value);
+					}
 				} else {
 					if ($filter instanceof FilterText) {
 						$this->applyFilterText($filter);


emololftw:
related to: https://github.com/contributte/datagrid/pull/298

radimvaculik:
@paveljanda 👀

paveljanda:
This really imho does not make any sense. When there is some spetial behaviour in Nextras, in should be handled in Nextras ano not in the parent abstract class.

radimvaculik:
@emololftw Ping.

emololftw:
@radimvaculik @paveljanda **I'm so sorry for my late response.** Of course, you're right - this specific behavior is useless to keep in the general class. Would you be open to accepting it if I implement it in the Nextras class case instead?

radimvaculik:
@emololftw Please.

PR #1202, opened 2025-08-06
Fix export bug of missing config files

### LINKED ISSUE #1201: Package without package.json and typescript config
The current Datagrid 7.1 package doesn't include Typescript config files. Is this behavior intentional, that the files are no longer included?

skyfly18:
fixed #1201

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -11,6 +11,3 @@ README.md export-ignore
 ruleset.xml export-ignore
 # Assets
 .npmignore export-ignore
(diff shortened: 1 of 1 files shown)


skyfly18:
package.json because Dependency Management.

f3l1x:
And others?

f3l1x:
Why do you need these files?

rollup.config.js export-ignore
tsconfig.json export-ignore

PR #1203, opened 2025-08-08
Example of nested components when setting sortable handler in docs

pilarik31:
I've added a mention of nested component usage in sortable handler to the docs.


radimvaculik:
LGTM.

Issue #1208, opened 2025-08-30
URL refreshing is not working


llsm:
Hi, 
i installed latest version and URL refreshing stopped working when i use filtering or change items per page. It works only when changing pages via buttons. 

Any ideas?

Thanks

radimvaculik:
@llsm Could you test my changes, please?

llsm:
Yes, looks better now. Thanks

PR #1236, opened 2026-01-05
Fix URL parameter handling for filtering and perPage changes (issue #…

radimvaculik:
…1208)

Fixes issue #1208 where URL refreshing stopped working when using filtering or changing items per page. Only pagination via buttons was working.

The problem was that the filterSucceeded() method was adding ALL form filter values to the persistent $filter parameter, including empty ones. This caused URLs to include empty filter parameters like ?grid-filter[name]=, matching the reported regression from version 6.x.

Changes:
- Only non-empty filter values are now added to the persistent $filter parameter
- Empty filter values are explicitly unset from $filter
- The $hasNonEmptyFilter flag is now used to determine if filters are active, instead of just checking if values exist
- Empty filter values remain stored in session storage for recall, but don't appear in URL parameters

This ensures clean URLs and maintains feature parity with version 6.x.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/Datagrid.php
+++ b/src/Datagrid.php
@@ -1383,7 +1383,14 @@ class Datagrid extends Control
 		}
 
 		$storedFilters = $this->getStorageData('_grid_filters', []);
+		$hasNonEmptyFilter = false;
+
 		foreach ($values as $key => $value) {
+			/**
+			 * Check if value is empty
+			 */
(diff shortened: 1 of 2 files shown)


f3l1x (approved):
LGTM

PR #1246, opened 2026-02-08
Extract reset filter button into reusable snippet

radimvaculik:
For outer filter, make reset filter button closer.

Screenshot 2026-02-08 at 15 30 13


radimvaculik:
@f3l1x Check it out.

PR #1254, opened 2026-02-22
Fix: ColumnsSummary shows incorrect totals after inline edit

### LINKED ISSUE #579: Summary reload after editing item
After editing item there is only 1 row (the edited one) loaded. Summary is then computed from this row only.

Can replicate on example grid https://ublaboo.org/datagrid/.

1. hit edit on any row
2. then save (no need to change anything)
3. look at the summary below

radimvaculik:
When redrawing a single row, ColumnsSummary was only populated with that one row's values instead of all visible rows. Fixed by using filterData (instead of filterRow) when columnsSummary is present, so all rows are iterated and added to the summary, while only the target row is rendered.

Closes #579.

### DIFF
--- a/src/Datagrid.php
+++ b/src/Datagrid.php
@@ -370,16 +370,35 @@ class Datagrid extends Control
 		 */
 		$rows = [];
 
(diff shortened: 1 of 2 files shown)


radimvaculik:
@f3l1x Covered.

PR #1274, opened 2026-03-17
Fix: MSSQL text filter double N-prefix (NN'...') with Dibi 4.1+

### LINKED ISSUE #908: Adding NN to query with SqlsrvDriver (Dibi\Fluent)
### Specification
When specifying string value in query, datagrid adds two "N" before leading string apostrophe.
For example:
`$fluent->where("type = 'type1'")` datagrid represents as `"WHERE type = NN'type1'"`
### Source
I guess the reason is double calling of method escapeText() from SqlsrvDriver. Dibi from version 4.1 adds "N" before escaped text. Calling once is OK, calling twice doubles leading "N", which MSSQL server doesn't understand.
### Workaround
- delete "N" from this method in SqlsrvDriver.php or
- use Dibi 4.0.3 or older

radimvaculik:
## Summary

Fixes #908

The `DibiFluentMssqlDataSource::applyFilterText()` override was broken in multiple ways:

- **Double N-prefix**: Used `%s` / direct string interpolation in a way that caused `SqlsrvDriver` (Dibi 4.1+) to call `escapeText()` twice, producing `NN'value'` instead of `N'value'`
- **SQL injection**: LIKE search used `sprintf('%s LIKE "%%%s%%"', $column, $value)` — directly interpolating user input into SQL
- **Wrong string delimiters**: Used double quotes (`"..."`) which are identifier delimiters in MSSQL, not string delimiters
- **Missing split-words search**: Ignored `$filter->hasSplitWordsSearch()`

## Fix

Remove the broken override entirely. The base class `DibiFluentDataSource::applyFilterText()` handles all cases correctly via Dibi's `%~like~` modifier, which calls the driver's escape method exactly once (producing correct `N'...'` on MSSQL).

### DIFF
--- a/src/DataSource/DibiFluentMssqlDataSource.php
+++ b/src/DataSource/DibiFluentMssqlDataSource.php
@@ -5,10 +5,8 @@ namespace Contributte\Datagrid\DataSource;
 use Contributte\Datagrid\Exception\DatagridDateTimeHelperException;
 use Contributte\Datagrid\Filter\FilterDate;
 use Contributte\Datagrid\Filter\FilterDateRange;
-use Contributte\Datagrid\Filter\FilterText;
 use Contributte\Datagrid\Utils\DateTimeHelper;
 use Dibi\Fluent;
-use Dibi\Helpers;
 use Dibi\Result;
 use UnexpectedValueException;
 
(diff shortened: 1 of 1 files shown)


radimvaculik:
@f3l1x Hard to setup all new MSSQL env.

f3l1x:
It's not needed to setup MSSQL, you can compare SQL queries or something like that.

TumaMilan (approved):
Fix looks OK, tested on my config. 
Thank you very much!

PR #1282, opened 2026-04-16
Fix: FilterSelect implicit validation breaks form on hidden column (#1281)

### LINKED ISSUE #1281: getValues() on filter container triggers E_USER_WARNING when a FilterSelect column is hidden
## Summary

  When a datagrid column with `setFilterSelect(...)` is hidden (via `setDefaultHide(true)` or runtime `handleHideColumn`), submitting any filter triggers a PHP `E_USER_WARNING`:

  > `Nette\Forms\Container::getValues() invoked but the form is not valid (form 'filter').`

  Source: `Datagrid::filterSucceeded()` at `src/Datagrid.php:1383`:

  ```php
  $values = $form->getComponent('filter')->getValues();
  ```

  Versions: `contributte/datagrid` 7.2.0, `nette/forms` 3.2.8, PHP 8.5.

  ## Reproduction

  Grid needs at least one **hidden** column with `setFilterSelect(...)` plus any other visible filter to trigger a submission:

  ```php
  $grid->addColumnText('name', 'Name')
      ->setFilterText();

  $grid->addColumnText('status', 'Status')
      ->setDefaultHide(true)
      ->setFilterSelect(['yes' => 'Yes', 'no' => 'No']);
  ```

  Steps:
  1. Load the grid (`status` column is hidden by default).
  2. Type anything into the `name` filter and submit.
  3. Observe the warning in the Tracy bar / error log.

  Only `FilterSelect` triggers this — `FilterText`, `FilterMultiSelect`, `FilterDate`, `FilterRange` don't.

  ## Notes

  `nette/forms`' `Container::getValues()` emits `E_USER_WARNING` when called on an invalid container. The filter container becomes invalid because `SelectBox` adds an implicit validation rule in its
  constructor (requiring a selected value unless a prompt is set), and a hidden column's select submits no value — so the rule fails.

radimvaculik:
## Summary

Fixes #1281.

`Nette\Forms\Controls\SelectBox` constructor adds an implicit `Filled` validation rule when no prompt is set. For a `FilterSelect` on a column that's hidden via `setDefaultHide()` or `handleHideColumn`, the input isn't rendered, so nothing is submitted for it. The rule then fails, the filter container becomes invalid, and `Container::getValues()` in `Datagrid::filterSucceeded()` triggers:

> `Nette\Forms\Container::getValues() invoked but the form is not valid (form 'filter').`

Filters are search inputs, not required data entry, so this PR resets the rules on the underlying `SelectBox` in `FilterSelect::addControl()`. The fix is inherited by `FilterMultiSelect` (no-op there since `addMultiSelect` doesn't add implicit validation).

## Test plan

- [x] Added regression test `testFilterSelectHasNoImplicitValidation` in `tests/Cases/FilterTest.phpt` — verified it fails before the fix and passes after.
- [x] `make tests` — 29 passed, 1 skipped (MySQL not running).
- [x] `make phpstan` — clean.
- [x] `make cs` — clean.

@jaroslavlibal could you give this branch a try in your project and confirm the warning is gone? 🙏

### DIFF
--- a/src/Filter/FilterSelect.php
+++ b/src/Filter/FilterSelect.php
@@ -104,6 +104,13 @@ class FilterSelect extends OneColumnFilter
 	{
 		$input = $container->addSelect($key, $name, $options);
 
+		// SelectBox adds an implicit "Filled" rule when no prompt is set.
+		// A filter is never required, and when its column is hidden the input
+		// isn't rendered, so nothing is submitted and the rule fails — which
+		// invalidates the whole filter container and triggers a warning on
+		// getValues(). Filters don't need validation, so drop it here.
+		$input->getRules()->reset();
+
 		if ($this->getPrompt() !== null) {
 			$input->setPrompt($this->getPrompt());
 		}
--- a/tests/Cases/FilterTest.phpt
+++ b/tests/Cases/FilterTest.phpt
@@ -110,6 +110,40 @@ final class FilterTest extends TestCase
 		Assert::count(0, $grid->filter);
 	}
 
+	/**
+	 * Regression test for https://github.com/contributte/datagrid/issues/1281
+	 *
+	 * A FilterSelect on a hidden column isn't rendered in HTML, so no value is
+	 * submitted for it. SelectBox's constructor adds an implicit "Filled" rule
+	 * when no prompt is set, which used to fail validation for the missing
+	 * value, invalidate the filter container, and trigger an E_USER_WARNING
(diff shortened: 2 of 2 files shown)


jaroslavlibal:
I can confirm that the fix works in my project. Thank you!

jaroslavlibal:
Is there anything else I can do to get this into the master branch soon, please?

radimvaculik:
@f3l1x Could you review and merge it, please?

PR #1283, opened 2026-05-05
Custom where condition for NetteDatabaseDataSource

vcernik:
Adds custom where condition for NetteDatabaseDataSource

### DIFF
--- a/.docs/filters.md
+++ b/.docs/filters.md
@@ -90,6 +90,16 @@ $grid->addFilterText('custom', 'Custom search:', 'name')
 	});
 ```
 
+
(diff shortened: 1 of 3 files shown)


radimvaculik:
LGTM. But seems like breaking change.

vcernik:
Yes, it seems. But when previous `getDataSource` returns `Explorer`, it's completely useless - there is no logical connection to object NetteDatabaseDataSource.

_I hope I'm making myself clear. I'm excited - this is my first PR :D Thank you for this awesome Datagrid! Hezký den ;)_

PR #1298, opened 2026-07-27
Fix url refresh after filter

llsm:
The latest version does not update the URL in the browser after filtering. This change fixes it. 

URL refresh isn't working in my project, nor in the demo at https://examples.contributte.org/datagrid-skeleton/.


radimvaculik (commented):
Thanks for digging into this — the diagnosis is correct.

Historically the URL refresh was done in JS: the old `assets/datagrid.js` had a `datagrid.url` extension that reacted to `payload._datagrid_url`, rebuilt the URL from `payload.state` and called `history.replaceState()`. That extension was lost in the TS rewrite (dca24b4) — today `_datagrid_url` only exists in `assets/types/ajax.d.ts`, nothing consumes it. Hence the regression.

Delegating to Naja is the right call (`HistoryHandler.pushNewState()` does `if (payload.postGet && payload.url) options.href = payload.url`), and it matches what `handleEdit()` already does. Three things I'd like to resolve before merging:

### 1. It breaks `setRefreshUrl(false)`

The new lines are unconditional, so `$this->refreshURL` becomes meaningless (nothing reads `_datagrid_url` anymore). It needs to be guarded:

```php
$this->getPresenterInstance()->payload->_datagrid_url = $this->refreshURL;
$this->getPresenterInstance()->payload->_datagrid_name = $this->getFullName();

if ($this->refreshURL) {
    $this->getPresenterInstance()->payload->postGet = true;
    $this->getPresenterInstance()->payload->url = $this->link('this');
}
```

### 2. Only one of the code paths is fixed

`_datagrid_url` is still sent (uselessly) from four other places:

- `reloadTheWholeGrid()` — used by `handleSort()`, `handleResetFilter()`, `handleResetColumnFilter()`. This one is worse than "URL not updated": those are GET ajax links, so Naja pushes the **signal URL** (`?do=grid-sort&…`) into history.
- `handleGetChildren()`
- `redrawItem()`
- `handleRefreshState()` — a method whose only purpose is "Simply refresh url", currently a no-op.

Could you extract this into a small private helper and call it from all of them, so it doesn't drift apart again?

### 3. Behavioural change: `pushState` instead of `replaceState`

The old JS used `replaceState`. Naja with the default `history: true` does `pushState`, so with filter autosubmit (debounced, `assets/plugins/features/autosubmit.ts`) every few keystrokes adds a history entry and the Back button walks through intermediate filter states. One-line fix in `src/templates/datagrid.latte`:

```latte
{form filter, class => 'ajax', data-naja-history => 'replace', data-naja-unique => "datagrid-filter-{$control->getFullName()}"}
```

---

Minor: `handleEdit()` already sets `postGet`/`url` unconditionally, i.e. it ignores `refreshURL` too — worth unifying via the same helper. The failing `codecov/patch` check isn't a blocker here.

